Issue #2, opened 2016-06-01
Swappable context API


eirslett:
We're currently working on a refactor of zipkin-js, where you, instead of using continuation-local-storage, can provide your own context API. This is handy for people who either don't want to use continuation-local-storage for whatever reasons, and/or people who already have a context object they pass around explicitly in their application code.
Another use case would be to run zipkin-js in the browser, where we can inject zone.js as a context api.
The work is going on in a feature branch.

PR #5, opened 2016-06-02
Feature/context

sveisvei:
Closes https://github.com/openzipkin/zipkin-js/issues/2


packages/zipkin-context-cls/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "zipkin-context-cls",
+  "version": "0.1.9",
+  "description": "A Context API implementation that uses continuation-local-storage under the hood",
+  "main": "src/CLSContext.js",
+  "scripts": {
+    "test": "node ../../node_modules/mocha/bin/mocha --require ../../test/helper.js"
+  },
+  "author": "Openzipkin <openzipkin.alt@gmail.com>",
```
adriancole: kill me for mentioning the case format is OpenZipkin? :P

packages/zipkin/src/batch-recorder.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  _writeSpan(id) {
+    const spanToWrite = this.partialSpans.get(id);
+    // ready for garbage collection
+    this.partialSpans.delete(id);
+    this.logger.logSpan(spanToWrite);
+  }
+
+  _updateSpanMap(id, updater) {
```
adriancole: heh looks like finagle!

packages/zipkin/src/batch-recorder.js
```diff
@@ -0,0 +1,118 @@
...
+          this._annotate(span, rec, thriftTypes.SERVER_RECV);
+          break;
+        case 'Message':
+          this._annotate(span, rec, rec.annotation.message);
+          break;
+        case 'Rpc':
+          span.setName(rec.annotation.name);
+          break;
+        case 'ServiceName':
+          span.setServiceName(rec.annotation.serviceName);
```
adriancole: generally, this is held constant per app, but I don't know the scope of this (and also recognise this is emulating finagle's behavior)

packages/zipkin/src/tracer/record.js
```diff
@@ -0,0 +1,13 @@
+class Record {
```
adriancole: for types that are basically the same as finagle, should we link? or cover in overall readme that this is largely inspired by finagle's Tracer

packages/zipkin-context-cls/README.md
```diff
@@ -0,0 +1,19 @@
...
+## Usage:
+
+```javascript
+const CLSContext = require('zipkin-context-cls');
+const tracer = new Tracer({
+  ctxImpl: new CLSContext('zipkin')
+});
+```
+
+## A not on CLS context vs. explicit context
```
adriancole: typo "A note"

adriancole:
looks great! request.. could we add a sample of an in-browser trace creation? well.. it might imply we have an http reporter to be realistic (unless scribe works in-browser)

packages/zipkin-context-cls/package.json
eirslett: My fault; it's Openzipkin all over zipkin-js now. We can just do a search-replace later on.

packages/zipkin/src/batch-recorder.js
eirslett: Yes, it was actually ported (more or less) directly from Finagle to JavaScript. That's why there's even an option/some/none inplementation in there. We should make zipkin-js more JavaScript-idiomatic later on.

packages/zipkin/src/tracer/record.js
eirslett: Could be an idea, but I imagine the implementations will diverge somewhat in the longer run.

eirslett:
> looks great! request.. could we add a sample of an in-browser trace creation? well.. it might imply we have an http reporter to be realistic (unless scribe works in-browser)

I'll throw together an isomorphic http transport that can be used both from the browser and from node.js. And a context implementation with zone.js. That will be enough to create an in-browser example. (Strictly speaking, we could probably use ExplicitContext in the browser as well, but overall I imagine zone.js will be prettier)

packages/zipkin/src/batch-recorder.js
eirslett: The ServiceName here (as opposed to LocalAddress) is the remote service name. (e.g. "mysql")

packages/zipkin/src/tracer/index.js
```diff
@@ -0,0 +1,126 @@
+require('colors');
```
sveisvei: oups, will remove this debugging.

packages/zipkin/src/batch-recorder.js
adriancole: ahh so this ends up being the "sa" binary annotation. gotcha

PR #8, opened 2016-06-07
fixed typo in expressMiddleware.js

iyersan:


eirslett:
Ouch! Thanks for finding that one!
@sveisvei we need to implement this test soon! :wink: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-express/test/integrationTest.js#L8

eirslett:
Released as v0.2.1!

sveisvei:
💯

PR #10, opened 2016-06-08
update README.md to working example

hakanson:
Adding .expressMiddleware to require() prevents the following error:

app.use(zipkinMiddleware({
        ^

TypeError: zipkinMiddleware is not a function


eirslett:
Thank you!

PR #20, opened 2016-07-06
Add Restify instrumentation package

jquatier:
I created a port of the express middleware that works with [Restify](http://restify.com/). There's just a few APIs that vary between Restify and Express, mainly around what methods are available on the request object. I updated the integration tests to match and they currently pass. The README contains example usage as well. 

Anyway, I'm open to any changes. I could also publish this myself but thought it might belong with the other instrumentations you have available. Let me know! Thanks!

### DIFF
--- a/packages/zipkin-instrumentation-restify/package.json
+++ b/packages/zipkin-instrumentation-restify/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "zipkin-instrumentation-restify",
+  "version": "0.2.2",
+  "description": "Restify middleware for instrumentation with Zipkin.js",
+  "main": "index.js",
+  "scripts": {
+    "test": "node ../../node_modules/mocha/bin/mocha --require ../../test/helper.js"
+  },
+  "author": "OpenZipkin <openzipkin.alt@gmail.com>",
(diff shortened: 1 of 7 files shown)


packages/zipkin-instrumentation-restify/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "zipkin-instrumentation-restify",
+  "version": "0.1.0",
```
eirslett: verion is 0.2.2 - all the packages in zipkin-js have the same version.

eirslett:
Thanks a lot for the PR!

jquatier:
@eirslett Done. I also updated the top-level README which has the list of packages. I can also put in a PR for the [instrumentations page](https://github.com/openzipkin/openzipkin.github.io/blob/master/_data/existing_instrumentations.yml) on zipkin.io once this package is published.

eirslett:
Thanks a lot!

jquatier:
Great! Let me know when you think this will be available on NPM. Wasn't sure how often you run lerna-publish. In the meantime, I'll update the other repo with a PR.

PR #23, opened 2016-07-14
Adding HTTP transport module

### LINKED ISSUE #15: equivalent of zipkin-java JsonCodec?
zipkin-java has a [JsonCodec](https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/internal/JsonCodec.java) class that is used by things like [ElasticsearchSpanConsumer](https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java) or spring-cloud-sleuth [HttpZipkinSpanReporter](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java).  Can support for JsonCodec be added?

I am looking to leverage the "Spring Cloud Zipkin UI" (not sure of the proper name, but screenshots at https://spring.io/blog/2016/02/15/distributed-tracing-with-spring-cloud-sleuth-and-spring-cloud-zipkin) and this seems like a required prerequisite.

jquatier:
Fixes #15. I modeled most of this after the scribe and kafka loggers. Basically, there is just an internal queue that gets checked every second for new spans. If there are spans, it batches them up in a single JSON post (an array) and sends them to a configurable HTTP endpoint.

Let me know if you'd like any changes made! I've tested this on a Restify-based service and I also added unit / integration tests.

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -1,0 +1,43 @@
+/* eslint-disable no-console */
+const fetch = require('node-fetch');
+
+class HttpLogger {
+  constructor({endpoint, httpInterval = 1000}) {
+    this.endpoint = endpoint;
+    this.queue = [];
+
+    const timer = setInterval(() => {
+      this.processQueue();
+    }, httpInterval);
+    timer.unref();
+  }
+
+  logSpan(span) {
+    this.queue.push(span.toJSON());
+  }
+
+  processQueue() {
+    if (this.queue.length > 0) {
+      const postBody = JSON.stringify(this.queue);
+      fetch(this.endpoint, {
+        method: 'POST',
+        body: postBody,
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json'
+        }
+      }).then((response) => {
+        if (response.status !== 202) {
+          console.error('Unexpected response while sending Zipkin data, status:' +
+            `${response.status}, body: ${postBody}`);
+        }
+        this.queue.length = 0;
+      }).catch((error) => {
+        console.error('Error sending Zipkin data', error);
+        this.queue.length = 0;
+      });
+    }
+  }
+}
+
+module.exports = HttpLogger;
--- a/packages/zipkin-transport-http/test/integrationTest.js
+++ b/packages/zipkin-transport-http/test/integrationTest.js
@@ -1,0 +1,39 @@
+const {Tracer, BatchRecorder, Annotation, ExplicitContext} = require('zipkin');
+const HttpLogger = require('../src/HttpLogger');
+const express = require('express');
+const bodyParser = require('body-parser');
+
+describe('HTTP transport - integration test', () => {
+  it('should send trace data via HTTP', function(done) {
+    const app = express();
+    app.use(bodyParser.json());
+    app.post('/api/v1/spans', (req, res) => {
+      res.status(202).json({});
+      const traceData = req.body;
+      expect(traceData.length).to.equal(1);
+      expect(traceData[0].name).to.equal('GET');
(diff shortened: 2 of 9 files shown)


jquatier:
I made the version of this package 0.2.3, but a new version of the zipkin core package will need to be published too so I ticked that to 0.2.3. Let me know if I should do that differently. Thanks!

jquatier:
I ticked the version number in the `lerna.json` file to `0.2.3` since there are changes in the core zipkin module. One thing I noticed is that the restify instrumentation that I published is already on `0.2.3` in NPM, but in github its still `0.2.2`. Anyway, let me know if I did that right.

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -0,0 +1,41 @@
...
+      fetch(this.endpoint, {
+        method: 'POST',
+        body: postBody,
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json'
+        }
+      }).then((response) => {
+        if (response.status === 202) {
+          this.queue.length = 0;
```
eirslett: What do you think about truncating the queue in every case, regardless of whether the request succeeds or not? We don't require consistency in Zipkin trace data. My concern is that if the zipkin server is down, the posting would fail, and this.queue would never be emptied, leading to a memory leak.

packages/zipkin-transport-http/test/integrationTest.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-console */
```
eirslett: I don't think this is needed? I cannot see any console.log statements in this file, have I missed one?

packages/zipkin/src/internalRepresentations.js
```diff
@@ -15,6 +15,16 @@ Endpoint.prototype.toThrift = function toThrift() {
     port: this.port
   });
 };
+Endpoint.prototype.toJSON = function toJSON() {
+  return {
+    ipv4: this.formatIPv4(this.host),
+    port: this.port
+  };
+};
+Endpoint.prototype.formatIPv4 = function formatIPv4(host) {
```
eirslett: This function doesn't need to be on Endpoint.prototype, it can live on its own. (It doesn't need access to `this`)

eirslett:
Great work!

adriancole:
:clap:

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable no-console */
+const fetch = require('node-fetch');
+
+class HttpLogger {
+  constructor({endpoint, httpInterval = 1000}) {
+    this.endpoint = endpoint;
+    this.queue = [];
+
+    setInterval(() => {
```
sveisvei: Maybe unref this? https://nodejs.org/api/timers.html#timers_timeout_unref

sveisvei:
👍  lgtm

packages/zipkin-transport-http/src/HttpLogger.js
jquatier: Good call. I'll take care of this and the other items today. Thanks!

packages/zipkin-transport-http/src/HttpLogger.js
jquatier: That's a good point! I'll make the change.

jquatier:
All feedback has been addressed, and it looks like the travis build is back so 👍

eirslett:
Thanks a lot!

PR #25, opened 2016-07-19
unref the timer in BatchRecorder

jquatier:
I noticed my node process wasn't exiting when there was nothing else running, and discovered it was the timer in BatchRecorder that was still running. Anyway, made the same fix that @sveisvei suggested to BatchRecorder to solve the problem.


sveisvei:
Nice! Looks good to merge and patch-release. _(on the go, will do later if noone else beats me to it)_

adriancole:
I'll try!

adriancole:
0.2.5 on the way! thx for the fix

PR #26, opened 2016-07-20
fix "Missing required argument recorder" error

hakanson:
Rename httpRecorder to recorder because Tracer is specifically looking for a parameter named "recorder" or it throws an error like:

Error: Tracer: Missing required argument recorder.
    at requiredArg (/Users/Kevin/myservice1/node_modules/zipkin/src/tracer/index.js:11:9)
    at new Tracer (/Users/Kevin/myservice1/node_modules/zipkin/src/tracer/index.js:17:16)
    at Object.<anonymous> (/Users/Kevin/myservice1/app.js:17:16)


jquatier:
Ah, Yep. Nice find! Looks good to me.

PR #31, opened 2016-08-14
Guards timer.unref for browsers who might not have it

### LINKED ISSUE #30: Make BatchRecorder browser-friendly?
I'm trying to get zipkin-js working in the browser, by browserify'ing the following

It eventually dies on `timer.unref();` from BatchRecorder. I'm not sure if there are tricks to make it not die.

Here's my example code:
```
const {Tracer, BatchRecorder, ExplicitContext} = require('zipkin');
const {HttpLogger} = require('zipkin-transport-http');

const recorder = new BatchRecorder({
  logger: new HttpLogger({
    endpoint: 'http://localhost:9411/api/v1/spans'
  })
});
const ctxImpl = new ExplicitContext();
const tracer = new Tracer({ctxImpl, recorder});

const zipkinFetch = wrapFetch(fetch, {tracer, serviceName: 'browser'});

var result = fetch('/api')
result.then(function(response) {
  console.log('response', response)
  console.log('header', response.headers.get('Content-Type'))
  return response.text()
}).then(function(text) {
  console.log('got text', text)
}).catch(function(ex) {
  console.log('failed', ex)
})
```

adriancole:
Fixes #30

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -9,7 +9,9 @@ class HttpLogger {
     const timer = setInterval(() => {
       this.processQueue();
     }, httpInterval);
-    timer.unref();
+    if (timer.unref) { // unref might not be available in browsers
+      timer.unref(); // Allows Node to terminate instead of blocking on timer
+    }
   }
 
   logSpan(span) {
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -25,7 +25,9 @@ class BatchRecorder {
         }
       });
     }, 1000);
-    timer.unref();
+    if (timer.unref) { // unref might not be available in browsers
+      timer.unref(); // Allows Node to terminate instead of blocking on timer
+    }
   }
 
   _writeSpan(id) {


jquatier:
Looks good! 👍

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -9,7 +9,9 @@ class HttpLogger {
     const timer = setInterval(() => {
       this.processQueue();
     }, httpInterval);
-    timer.unref();
+    if (timer.unref) {
+      timer.unref();
```
adriancole: @jquatier ps what's the impact of "not doing this" in javascript that doesn't support it? I should make a code comment I think.

jquatier: I think a comment is good just so we can point out why we are doing it. `timer.unref()` really only applies to Node where there's a running process that won't exit until the event loop is done. From a browser, there's still an event loop but the window doesn't close when the event loop is done processing events. Anyway, just one of those weird things about writing universal JS. 👍

adriancole: thx updated comments

Issue #33, opened 2016-08-17
HTTP transporter does not send open records when node process is about to terminate


ocastx:
The current implementation of the HTTP transporter does not send open records in it's queue when the node process finished executing (no events left in the queue). 

``` javascript
if (timer.unref) { // unref might not be available in browsers
     timer.unref(); // Allows Node to terminate instead of blocking on timer
}
```

Why shouldn't the transporter send all open records before exiting?

jquatier:
The reason we `unref` the timer is so the process exits when there's nothing else running in the event loop, otherwise your process would continue running forever. There's a `process.on('exit')` hook that could be used, but you can only do synchronous operations there, so that wouldn't work. 

Anyway, would like to hear if anyone has a good solution. I do remember @eirslett telling me that high consistency isn't a requirement for Zipkin trace data, so I think that's why we might assume that if a few spans are lost as the server is coming down, it's an acceptable scenario. The problem of course is if your Node script just runs for a second or 2 from start to finish, you'll never get anything unless you wait for the recorder to run. You can set the `httpInterval` to something much smaller than a second if needed.

ocastx:
I would suggest adding a setTimeout with the same `httpInterval` which is overwritten every time a new span is put into the queue. With this the node process would wait the minimum time so the interval fires. Also if the queues is empty and the timeout is still running it could then be terminated so the process would not shutdown gracefully with a delay of like one sec.

I already built this to fix my problem, could create a PR for this if wanted.

jquatier:
That sounds like a good solution. I think a similar change could be made to the BatchRecorder since it also uses a timer. Hopefully @sveisvei or @adriancole can comment on this, too.

jthomas:
I also hit this issue during development of a [new instrumentation](https://github.com/jthomas/zipkin-instrumentation-openwhisk) for a serverless function platform.

Serverless platforms spawn a new process for each invocation. Once the function handler returns, the process is terminated. This means the final spans often haven't finished sending before the process exits.

I [worked around](https://github.com/jthomas/zipkin-instrumentation-openwhisk/blob/master/lib/func.js#L39-L44) this issue but would prefer a custom event or configuration flag to modify this behaviour in the HTTP logger.

adriancole:
most instrumentation I've noticed attempt to flush on close, although not always via a shutdown hook. For example, in the java side, up to 1 second of blocking is permitted before spans are dropped on close

adriancole:
[...]

Issue #3, opened 2016-06-01
Time in microseconds


eirslett:
The current implementation in time.js has a bug:

```javascript
const d = new Date();
return d.getTime() * 1000
  + d.getMilliseconds(); // <-- This should be microseconds, not milliseconds
```

We want to get the number of *microseconds* since 1970, not the number of milliseconds. That's why we take d.getTime(), which returns number of milliseconds from 1970, and multiply by 1000, to get number of microseconds since 1970.

In order to solve this, we need to find a timer with microsecond precision, that preferably works both in the browser and in node.js.

sveisvei:
Using https://nodejs.org/api/process.html#process_process_hrtime_time `process.hrtime()` for node.js.

Browsers is another thing, and does not have this implemented yet. (https://www.w3.org/TR/hr-time/)
There is a shim we can use: https://github.com/kumavis/browser-process-hrtime

jdarling:
This might help, it uses process.hrtime if its available, if not it falls back to new Date.getTime

```js
const counter = (function(){
  if(process && process.hrtime){
    var getNanoSeconds = function(){
      var hr = process.hrtime();
      return hr[0] * 1e9 + hr[1];
    };
    var loadTime = getNanoSeconds();
    return function(from){
      var v = (getNanoSeconds()-loadTime) / 1e6;
      return from?v-from:v;
    };
  }
  return function(from){
    var v = (new Date()).getTime();
    return from?v-from:v;
  };
})();
```

This way if you have any values past the . you know you have higher precision values available.  If not then you still have the milli's that you can utilize.

adriancole:
Thanks for the help, folks. I might try something including these tips.

adriancole:
heh I just discovered that browser-process-hrtime isn't quite right as it returns relative seconds not epoch ones https://github.com/kumavis/browser-process-hrtime/issues/4 cc @basvanbeek

adriancole:
got a bit distracted today, but made heavy progress. I will have this done tomorrow

adriancole:
fix https://github.com/openzipkin/zipkin-js/pull/51

PR #34, opened 2016-08-17
Add timer to prevent node process to exit too early before sending spans

ocastx:
fixes openzipkin/zipkin-js#33

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -5,6 +5,7 @@ class HttpLogger {
   constructor({endpoint, httpInterval = 1000}) {
     this.endpoint = endpoint;
     this.queue = [];
+    this.httpInterval = httpInterval;
 
     const timer = setInterval(() => {
       this.processQueue();
@@ -16,6 +17,10 @@ class HttpLogger {
 
   logSpan(span) {
     this.queue.push(span.toJSON());
+
+    // Schedule a timeout so logs will also be processed
+    // when the node process is about to exit
+    this.timeoutTimer = setTimeout(() => {}, this.httpInterval);
   }
 
   processQueue() {
@@ -34,6 +39,11 @@ class HttpLogger {
             `${response.status}, body: ${postBody}`);
         }
         this.queue.length = 0;
+
+        // Clear the timeout timer when the queue is empty so
+        // if the node process is about to exit, it is not hold back by the timeout
+        clearTimeout(this.timeoutTimer);
+        this.timeoutTimer = null;
       }).catch((error) => {
         console.error('Error sending Zipkin data', error);
         this.queue.length = 0;
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -2,31 +2,34 @@
 const fetch = require('node-fetch');
 
 class HttpLogger {
-  constructor({endpoint, httpInterval = 1000}) {
+  constructor({endpoint, queueWait = 1000}) {
     this.endpoint = endpoint;
     this.queue = [];
-    this.httpInterval = httpInterval;
+    this.queueWait = queueWait;
 
-    const timer = setInterval(() => {
-      this.processQueue();
-    }, httpInterval);
-    if (timer.unref) { // unref might not be available in browsers
-      timer.unref(); // Allows Node to terminate instead of blocking on timer
-    }
+    process.on('uncaughtException', (err) => {
+      this.processQueue()
+        .then(() => {
+          console.log(err.stack);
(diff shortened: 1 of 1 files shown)


eirslett:
How would it look if we implemented process.on('exit', flushQueue) or something like that? It would be a little more explicit. My gut feeling is that registering an empty noop-timeout to fix the issue is a little brittle. E.g. one could imagine a JavaScript runtime engine discovered it was an empty callback/noop and it would "optimize" away the whole timeout, which would ruin this fix.

ocastx:
I get what you are saying, but I would suggest rather than using `process.on('exit', flushQueue)` to use a mechanism to schedule a timeout only if there are spans in the queue instead of using an interval at all. 

For example we could then schedule a timeout always when a new span is added to the queue, except if there is already a timer running. Then, if nothing is in the queue the timer would not be running and the interval would not use any cycles. Thereby we would not have a problem with a noop-timeout, an interval that is not executed when the process is terminated and have a more explicit way when the queue is about to be flushed.

jquatier:
I think that would definitely work. Unfortunately only synchronous operations can be put in process.on('exit'), so the process would still exit before the aync http execution ran.

adriancole:
Is there any way to automatically test this sort of behavior? (ex in the
java project, we have things that literally start/stop processes, though
not for this purpose)

eirslett:
> Unfortunately only synchronous operations can be put in process.on('exit')

Looks like `process.on('beforeExit', ...)` supports async calls, like for this use case.

> schedule a timeout only if there are spans in the queue instead of using an interval at all.

I guess in a shutdown hook, this could be implemented as a simple if-check? Only flush the queue if there are spans in there?

> have a more explicit way when the queue is about to be flushed.

Can you please explain a bit more how it would be more explicit? In my mind, adding a process shutdown hook is more explicit, since it's code that is supposed to be run when the process shuts down. But I don't have the complete picture in my mind...

Also, is this issue HTTP-specific, or does it also count for kafka and scribe? I would imagine that they suffer from the same problem, and that the fix would be similar. So it could maybe be refactored to work across all transports?

ocastx:
> Looks like process.on('beforeExit', ...) supports async calls, like for this use case.

Sounds like a good solution. Only problem with `beforeExit` is, that it's only triggered when the event loop is empty and the node process is about to exit. Therefore, when the process exits due to an exception the hook will not be triggered and the spans not send.
We can also use `process.on('uncaughtException', ...)` additionally to catch this case too.

> Also, is this issue HTTP-specific, or does it also count for kafka and scribe? I would imagine that they suffer from the same problem, and that the fix would be similar. So it could maybe be refactored to work across all transports?

Yes, I think the other transports would benefit from this refactor too.
I can update the PR if all can agree to use hooks to solve this.

eirslett:
> We can also use process.on('uncaughtException', ...) additionally to catch this case too.

I think uncaughtException only allows synchronous handlers? "The correct use of 'uncaughtException' is to perform synchronous cleanup of allocated resources (e.g. file descriptors, handles, etc) before shutting down the process." https://nodejs.org/api/process.html#process_event_uncaughtexception

A reasonable fallback for synchronous handling of uncaughtException would be to check if there are any spans in the queue that are not sent, and then console warn that x number of spans were not sent. Or something like that?

ocastx:
> A reasonable fallback for synchronous handling of uncaughtException would be to check if there are any spans in the queue that are not sent, and then console warn that x number of spans were not sent

Yes this would be a reasonable fallback but this wouldn't benefit data integrity. For the project I am currently working on this would not a be an appropriate solution because it would be necessary to deliver spans especially when the process is about to crash for better debugging.

jquatier:
Unfortunately, I don't think there's much you can do in the uncaught exception case. The process is going to exit. Even the solution in this PR doesn't solve that and I think logging as @eirslett suggested is appropriate. We could ecen consider logging the actual spans left in the queue, so that you still have the span data in the logs to look at perhaps?

adriancole:
ps BatchRecorder has a timeout bug caused by #3. Sending a fix shortly

adriancole:
ps I made a time-based test here https://github.com/openzipkin/zipkin-js/pull/36
looking around, the easier library to do this is https://github.com/vesln/timekeeper

ocastx:
You actually can do asynchronous operations in the uncaughtException hook. It is not recommend because the application is likely to be in an unknown state so you should not continue running it. 
We'll use this for our project to still send spans because this data is crucial but I can understand if you don't want this to be in the main project.

eirslett:
> but I can understand if you don't want this to be in the main project.

If it's possible, then it would probably be desirable if it can be done as safely as possible. Could maybe be a configuration flag?

ocastx:
I would prefer a configuration flag too if you guys are ok with it.

PR #35, opened 2016-08-19
Removes obsolete Annotation.duration

adriancole:
Annotation.duration is an obsolete field and was removed last year from
the thrift.


adriancole:
ps later, I can add Span.timestamp, duration which are supported, but not yet here

eirslett:
+1

adriancole:
thx for the review folks!

PR #36, opened 2016-08-19
Fixes bug where millis were used as micros

adriancole:
Before, the microsecond portion of the timestamp was filled with the
milliseconds value. This is mostly harmless, but incorrect. It is
better to truncate to millis until we record microsecond granularity.

See #3

### DIFF
--- a/packages/zipkin/src/time.js
+++ b/packages/zipkin/src/time.js
@@ -1,4 +1,4 @@
+// Returns the current time in epoch microseconds
 module.exports.now = function now() {
-  const d = new Date();
-  return d.getTime() * 1000 + d.getMilliseconds();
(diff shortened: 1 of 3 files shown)


eirslett:
I think we can use sinon for this (which we already use), it comes with lolex, with is similar to timekeeper.

adriancole:
cool. didn't know about that. will do!

adriancole:
I can't see how lolex can freeze the clock, yet.. still hunting

adriancole:
I mean freeze such that new Date() returns a constant

adriancole:
ahh got it working. nice tip!

adriancole:
ported to lolex removing sleeps! thx for the advice @eirslett!

PR #39, opened 2016-09-07
Fix javascript example in zipkin-transport-kafka package

sebastianvera:
There are two required keys when creating a Tracer, `ctxImpl` and `recorder`, the latter key was missing in the example. 

cc @rodrwan


eirslett:
Thanks!

PR #40, opened 2016-09-09
Fixed the reported path with querystring.

jpkrohling:
This fixes the reporting of the path when the path includes a query string. Previously, a URL like `http://localhost:8180/api/bonjour?_=1473407259102` would be reported on `http.url` as `http://localhost:8180/api/bonjour%3F_=1473407259102`. With this change, the `http.url` is now reported as `http://localhost:8180/api/bonjour?_=1473407259102`.

Before: https://snag.gy/xr4Zow.jpg
After: https://snag.gy/z4VB5O.jpg
References: https://nodejs.org/api/url.html#url_url_format_urlobject

### DIFF
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -23,6 +23,16 @@ function stringToIntOption(str) {
   }
 }
 
+function formatRequestUrl(req) {
+  const parsed = url.parse(req.originalUrl);
+  return url.format({
+    protocol: req.protocol,
+    host: req.get('host'),
+    pathname: parsed.pathname,
+    search: parsed.search
+  });
+}
+
 module.exports = function expressMiddleware({tracer, serviceName = 'unknown', port = 0}) {
   return function zipkinExpressMiddleware(req, res, next) {
     tracer.scoped(() => {
@@ -70,11 +80,7 @@ module.exports = function expressMiddleware({tracer, serviceName = 'unknown', po
 
       tracer.recordServiceName(serviceName);
       tracer.recordRpc(req.method);
-      tracer.recordBinary('http.url', url.format({
-        protocol: req.protocol,
-        host: req.get('host'),
-        pathname: req.originalUrl
-      }));
+      tracer.recordBinary('http.url', formatRequestUrl(req));
       tracer.recordAnnotation(new Annotation.ServerRecv());
       tracer.recordAnnotation(new Annotation.LocalAddr({port}));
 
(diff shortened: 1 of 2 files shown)


jpkrohling:
Are those checks expected to fail, or did I do something wrong? :)

packages/zipkin-instrumentation-express/src/expressMiddleware.js
```diff
@@ -70,10 +70,12 @@ module.exports = function expressMiddleware({tracer, serviceName = 'unknown', po
 
       tracer.recordServiceName(serviceName);
       tracer.recordRpc(req.method);
+      var parsed = url.parse(req.originalUrl);
```
adriancole: I think folks are preferring const

adriancole:
any chance you can add a unit test so that this doesn't later break?

adriancole:
ps thanks for the help!

eirslett:
It's failing because of the eslint code style check:

```
> node node_modules/eslint/bin/eslint.js packages
/home/travis/build/openzipkin/zipkin-js/packages/zipkin-instrumentation-express/src/expressMiddleware.js
  73:7  error  All 'var' declarations must be at the top of the function scope  vars-on-top
  73:7  error  Unexpected var, use let or const instead                         no-var
```

Basically, use const, not var.

packages/zipkin-instrumentation-express/src/expressMiddleware.js
eirslett: This is where eslint is failing the build; we only use const (or let).

jpkrohling:
Sorry, I somehow missed the message :) The PR has been updated, to use a `const` instead of `var`.

eirslett:
Is it possible to unit test it? Integration tests are usually slow...

jpkrohling:
Is there an example somewhere?

eirslett:
No, but I could make a sketch: make a function `formatRequestUrl(req) { /* business logic here */}` then make a unit test for that function. And then `tracer.recordBinary('http.url', formatRequestUrl(req));` in the code.

jpkrohling:
I see what you mean, but due to my limited knowledge of node/express in general, I couldn't add this simple unit test :) I did however implemented the changes in splitting the logic into a function, so, someone more skilled could probably easily add a unit test for that. The integration test is there for the moment, and can safely be removed once the unit test is added.

eirslett:
Ok, I see. That's perfectly understandable :)

jpkrohling:
Should the `appveyor` build have failed? I see this on the logs, which doesn't seem to relate to this change: 

    The install of memcached was NOT successful.
    Error while running 'C:\ProgramData\chocolatey\lib\memcached\tools\chocolateyInstall.ps1'.

sveisvei:
`appveyor` does not run at all, maybe we should remove and make `appveyor` support as an Issue / PR instead?

eirslett:
Yes, fixing AppVeyor should be a separate issue.

eirslett:
Anyways, let's merge this, we can revisit the unit test and the appveyor build later.

PR #47, opened 2016-10-03
Adds zipkin-instrumentation-redis support

### LINKED ISSUE #27: Feature request: zipkin-instrumentation-redis
Like https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-instrumentation-memcached, I suppose 😄

securingsincity:
Closes #27 

- Note I didn't include all of the redis commands, but a base set to get started similar to the memcached module

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ script:
 
 services:
   - memcached
+  - redis
 
 compiler: clang-3.6
 env:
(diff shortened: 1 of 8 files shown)


adriancole:
food for thought. in finagle, redis is a simple RPC trace with the span.name being the redis command.

so it would be something like:

span.name="append"
annotations="cs","cr" and possibly a ["Client Receive Error" pattern](https://github.com/openzipkin/openzipkin.github.io/issues/52)

https://github.com/twitter/finagle/blob/develop/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisTracingFilter.scala
https://github.com/twitter/finagle/blob/develop/finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/Command.scala

securingsincity:
@adriancole  that's super helpful I'll take a look thanks!

PR #48, opened 2016-10-04
upgrade kafka-node

### LINKED ISSUE #24: Upgrade kafka-node
This transitively pulls in vulnerable minimatch (https://nodesecurity.io/advisories/118). Any chance you could upgrade?

https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-transport-kafka

Doesn't seem like there's any breaking changes, should be straight forward: https://github.com/SOHU-Co/kafka-node/blob/master/CHANGELOG.md

sveisvei:
closes #24.

There should be only fixes here, so a minor jump is fine.
https://github.com/SOHU-Co/kafka-node/blob/master/CHANGELOG.md

### DIFF
--- a/packages/zipkin-transport-kafka/package.json
+++ b/packages/zipkin-transport-kafka/package.json
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "kafka-node": "^0.3.2"
+    "kafka-node": "^0.5.9"
   },
   "devDependencies": {
     "kafka-please": "^0.1.5",


packages/zipkin-transport-kafka/package.json
```diff
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "kafka-node": "^0.3.2"
+    "kafka-node": "^0.5.9"
```
SimenB: `"^0.5.0"`? Internally we depend on `0.5.3` locked (although that might be ok to bump)

adriancole:
on the note of releases, I noticed we've been bumping patch versions for things that seem like they should probably be minor versions. I know pre 1.0 versions don't say much, but should we start bumping minors for stuff like this? feels fair.

packages/zipkin-transport-kafka/package.json
SimenB: `"^0.5.0"`? Internally we depend on `0.5.7` locked (although that might be ok to bump)

sveisvei:
I feel this is a minor indeed. 👍

PR #50, opened 2016-10-08
Adds zipkin-instrumentation-redis support

### LINKED ISSUE #27: Feature request: zipkin-instrumentation-redis

securingsincity:
Closes #27

### DIFF
--- a/packages/zipkin-instrumentation-redis/src/zipkinClient.js
+++ b/packages/zipkin-instrumentation-redis/src/zipkinClient.js
@@ -14,7 +14,6 @@ module.exports = function zipkinClient(tracer, redis, options, serviceName = 're
     tracer.recordAnnotation(new Annotation.ClientSend());
     tracer.recordServiceName(serviceName);
     tracer.recordRpc(rpc);
-    tracer.recordBinary('redis.key', rpc);
   }
 
 
--- a/packages/zipkin-instrumentation-redis/src/zipkinClient.js
+++ b/packages/zipkin-instrumentation-redis/src/zipkinClient.js
@@ -1,0 +1,54 @@
+const {Annotation} = require('zipkin');
+const redisCommands = require('redis-commands');
+module.exports = function zipkinClient(tracer, redis, options, serviceName = 'redis') {
+  function mkZipkinCallback(callback, id) {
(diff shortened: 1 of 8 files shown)


packages/zipkin-instrumentation-redis/src/zipkinClient.js
```diff
@@ -0,0 +1,54 @@
...
+        tracer.recordAnnotation(new Annotation.ClientRecv());
+      });
+      callback.apply(this, args);
+    };
+  }
+  function commonAnnotations(rpc) {
+    tracer.recordAnnotation(new Annotation.ClientSend());
+    tracer.recordServiceName(serviceName);
+    tracer.recordRpc(rpc);
+    tracer.recordBinary('redis.key', rpc);
```
adriancole: shouldn't this be `redis.command`? Also, since it is the span name (which you can search on), do we need to log it again here?

eirslett: +1 we usually don't double-log the rpc/span name.

securingsincity: 👍  should be all fixed now. Completely agree it was unnecessary. I was following the memcached example a litlte too closely (copy pasta is never a good idea).

adriancole:
Thanks much

SimenB:
Awesome, thanks!

We actually went with [`ioredis`](https://github.com/luin/ioredis) as our client, because it supports Sentinels, Clusters and have promise support out of the box (it's the other one starred here: http://redis.io/clients#nodejs).

Would it be too much to ask to supply an integration for that lib as well? 👼  I'd think that a lot of the implementation could be shared

adriancole:
@SimenB in most cases, decisions like this rests mostly on interest in using and interest in maintaining. Are you keen to contribute support?

SimenB:
Not really... I could probably provide an initial implementation, at least (when we go through and try to add zipkin again, so not for a while)

adriancole:
shipped in v0.4.0

PR #51, opened 2016-10-12
Supports microsecond precision by employing process.hrtime

### LINKED ISSUE #3: Time in microseconds

### LINKED ISSUE #29: Support Span.timestamp/duration
Recent versions of the zipkin thrift have the fields Span.timestamp and duration. Particularly in the case of client spans, we can set an authoritative start, with duration accurate to microseconds. This would start by updating the model to the latest thrift.

https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift

adriancole:
This uses the same approach as Brave does, except using process.hrtime
instead of Java's `System.nanoTime`. Here's the approach in summary:

When creating a tracer or a span, save off the relative time and make
timestamps based on an high-precision offset from there.

In browsers, microsecond precision requires installing a shim like this:
```
// use higher-precision time than milliseconds
process.hrtime = require('browser-process-hrtime');
```

Here's an actual trace that uses this code, starting from the browser and continuing to the server:

screen shot 2016-10-12 at 12 49 54 pm

See https://nodejs.org/api/process.html#process_process_hrtime_time
See https://github.com/kumavis/browser-process-hrtime

Fixes #29
Fixes #3

### DIFF
--- a/packages/zipkin/src/internalRepresentations.js
+++ b/packages/zipkin/src/internalRepresentations.js
@@ -85,8 +85,8 @@ BinaryAnnotation.prototype.toJSON = function toJSON() {
 
 function MutableSpan(traceId) {
   this.traceId = traceId;
-  this.complete = false;
-  this.started = now();
+  this.startTimestamp = now();
+  this.startTick = process && process.hrtime ? process.hrtime() : undefined;
   this.name = None;
   this.service = None;
   this.endpoint = new Endpoint({});
@@ -100,11 +100,11 @@ MutableSpan.prototype.setServiceName = function setServiceName(name) {
   this.service = new Some(name);
 };
 MutableSpan.prototype.addAnnotation = function addAnnotation(ann) {
-  if (!this.complete && (
+  if (!this.endTimestamp && (
         ann.value === thriftTypes.CLIENT_RECV ||
         ann.value === thriftTypes.SERVER_SEND
         )) {
-    this.complete = true;
+    this.endTimestamp = now(this.startTimestamp, this.startTick);
   }
   this.annotations.push(ann);
 };
@@ -157,17 +157,25 @@ MutableSpan.prototype.toThrift = function toThrift() {
   return span;
 };
 MutableSpan.prototype.toJSON = function toJSON() {
+  let startedSpan = true;
   const trace = this.traceId;
   const spanJson = {
     traceId: trace.traceId,
     name: this.name.getOrElse('Unknown'),
-    id: trace.spanId,
-    timestamp: this.started
+    id: trace.spanId
   };
   trace._parentId.ifPresent(id => {
     spanJson.parentId = id;
   });
   spanJson.annotations = this.annotations.map(ann => {
+    // In the RPC span model, the client owns the timestamp and duration of the
+    // span. If we were propagated an id, we can assume that we shouldn't report
+    // timestamp or duration, rather let the client do that. Worst case we were
+    // propagated an unreported ID and Zipkin backfills timestamp and duration.
+    if (ann.value === thriftTypes.SERVER_RECV) {
+      // TODO: only set this to false when we know we in an existing trace
+      startedSpan = false;
+    }
     const annotationJson = ann.toJSON();
     this.overrideEndpointJSON(annotationJson);
     annotationJson.endpoint.serviceName = this.service.getOrElse('Unknown');
@@ -179,6 +187,12 @@ MutableSpan.prototype.toJSON = function toJSON() {
     annotationJson.endpoint.serviceName = this.service.getOrElse('Unknown');
     return annotationJson;
   });
+
+  // Log timestamp and duration if this tracer started and completed this span.
+  if (startedSpan && this.endTimestamp) {
+    spanJson.timestamp = this.startTimestamp;
+    spanJson.duration = Math.max(this.endTimestamp - this.startTimestamp, 1);
+  }
   return spanJson;
 };
 MutableSpan.prototype.toString = function toString() {
--- a/packages/zipkin/src/internalRepresentations.js
+++ b/packages/zipkin/src/internalRepresentations.js
@@ -1,5 +1,5 @@
 const thriftTypes = require('./gen-nodejs/zipkinCore_types');
-const {now} = require('./time');
+const {now, hrtime} = require('./time');
 const {Some, None} = require('./option');
 
 function Endpoint({host = 0, port = 0}) {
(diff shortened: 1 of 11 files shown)


packages/zipkin/src/time.js
```diff
@@ -1,4 +1,12 @@
 // Returns the current time in epoch microseconds
-module.exports.now = function now() {
-  return new Date().getTime() * 1000;
+// if startTimestamp and startTick are present, process.hrtime is used
+// See https://nodejs.org/api/process.html#process_process_hrtime_time
+module.exports.now = function now(startTimestamp, startTick) {
+  if (startTimestamp && startTick && process && process.hrtime) {
```
narqo: It feels that checking the existence of `process.hrtime` on every `now` call is redundant. You might do it once on module invocation and return different `now` implementations.

packages/zipkin/src/time.js
```diff
@@ -1,4 +1,12 @@
...
+    const hrtime = process.hrtime(startTick);
+    const elapsedMicros = Math.floor(hrtime[0] * 1000000 + hrtime[1] / 1000);
+    return startTimestamp + elapsedMicros;
+  } else {
+    return new Date().getTime() * 1000;
```
narqo: I believe you can use `Date.now()` as a much shorter equivalent of `new Date().getTime()`, don't you?

sveisvei: Date.now() is indeed in IE9 and later

packages/zipkin/src/internalRepresentations.js
```diff
@@ -85,8 +85,8 @@ BinaryAnnotation.prototype.toJSON = function toJSON() {
 
 function MutableSpan(traceId) {
   this.traceId = traceId;
-  this.complete = false;
-  this.started = now();
+  this.startTimestamp = now();
+  this.startTick = process && process.hrtime ? process.hrtime() : undefined;
```
sveisvei: As well could maybe put this in a function so it reads better.

```
function getHrTime() { 
    return process && process.hrtime ? process.hrtime() : undefined 
}
```

packages/zipkin/src/time.js
adriancole: takes me a very long time to figure out how to do things like this as I'm not very experienced in javascript. Can you give me an example of how to conditionally initialize now like you mentioned?

packages/zipkin/src/time.js
adriancole: Is it safe to assume we're IE9+ in general? if so, I'd also prefer Date.now()

packages/zipkin/src/internalRepresentations.js
adriancole: can do

packages/zipkin/src/time.js
sveisvei:
```js
const hrTimeSupport = typeof process !== 'undefined' && process.hrtime;

module.exports.now = hrTimeSupport ? nowHrTime : nowLegacy;

function nowLegacy () {
    return Date.now() * 1000;
} 

function nowHrTime(startTimestamp, startTick) {
    const hrtime = process.hrtime(startTick);
    const elapsedMicros = Math.floor(hrtime[0] * 1000000 + hrtime[1] / 1000);
    return startTimestamp + elapsedMicros;
};
```

packages/zipkin/src/time.js
adriancole: I tested this on the browser by using this..

```
// use higher-precision time than milliseconds
process.hrtime = require('browser-process-hrtime');
```

packages/zipkin/src/time.js
adriancole: the conditionals are needed, or we should rename the function as you need a seed timestamp since hrtime isn't based on wall clock time.

packages/zipkin/src/time.js
adriancole: done PTAL

packages/zipkin/src/internalRepresentations.js
adriancole: done

packages/zipkin/src/time.js
adriancole: done PTAL

packages/zipkin/src/tracer/index.js
```diff
@@ -21,6 +21,8 @@ class Tracer {
     this.sampler = sampler;
     this._ctxImpl = ctxImpl;
     this._defaultTraceId = this.createRootId();
+    this._startTimestamp = now();
+    this._startTick = process && process.hrtime ? process.hrtime() : undefined;
```
adriancole: oops.. @sveisvei this is why I used ternary, as I had it defined twice and seemed quicker as a one-line paste. should I define getHrTime again? or maybe add it to time.js?

packages/zipkin/src/time.js
```diff
@@ -1,4 +1,24 @@
...
+}
+
 // Returns the current time in epoch microseconds
-module.exports.now = function now() {
-  return new Date().getTime() * 1000;
-};
+// if startTimestamp and startTick are present, process.hrtime is used
+// See https://nodejs.org/api/process.html#process_process_hrtime_time
+module.exports.now = hrTimeSupport ? nowHrTime : nowLegacy;
+module.exports.hrtime = hrTimeSupport
```
adriancole: @sveisvei looks cleanest to me to keep hrtime here like this. wdyt?

packages/zipkin/src/tracer/index.js
adriancole: addressed in latest

packages/zipkin/src/time.js
sveisvei: This looks like its mighty fine. lgtm :)

adriancole:
thanks for the review @sveisvei and @narqo. The code's definitely better now.

adriancole:
fyi finished the example, which shows this working (from the browser) https://github.com/openzipkin/zipkin-js-example/pull/3

adriancole:
shipped in v0.4.0

PR #53, opened 2016-10-12
Adds remoteServiceName which maps to the ServerAddr binary annotation

### LINKED ISSUE #52: Correctly handle remote server address
I noticed in fetch and cujo examples like this:

```javascript
const nameOfRemoteService = 'youtube';
const zipkinFetch = wrapFetch(fetch, {tracer, serviceName: nameOfRemoteService});
```

In code, we aren't actually using the name as mentioned, as it is being used for the local service. In most libraries, we set the local service name (and address), as a part of initializing the tracer. 

When calling an uninstrumented server, like mysql or redis, we allow you to control the ServerAddr, which helps especially in the case of the dependency graph.

We should move declaring the local service to tracer initialization (at least a default one) and add support for ServerAddr, which examples like above could use.

adriancole:
The ServerAddr("sa") annotation is used to indicate the server. This is
used for uninstrumented clients, such as redis. When used properly, the
Zipkin dependencies graph will show those who call into such services.

This adds an option `remoteServiceName` which hints the name of the
peer. Where possible, other endpoint information such as host and port
are added.

Fixes #52

### DIFF
--- a/packages/zipkin-instrumentation-cujojs-rest/README.md
+++ b/packages/zipkin-instrumentation-cujojs-rest/README.md
@@ -12,7 +12,7 @@ const {restInterceptor} = require('zipkin-instrumentation-cujojs-rest');
 const tracer = new Tracer({ctxImpl, recorder}); // configure your tracer properly here
 
 const nameOfRemoteService = 'youtube';
-const client = rest.wrap(restInterceptor, {tracer, serviceName: nameOfRemoteService});
+const client = rest.wrap(restInterceptor, {tracer, remoteServiceName: nameOfRemoteService});
 
 // Your application code here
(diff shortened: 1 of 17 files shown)


adriancole:
tested manually, too, by making my demo app arbitrarily call memcached and then look at zipkin's dependency graph cc @eirslett @securingsincity

screen shot 2016-10-12 at 10 26 11 pm
screen shot 2016-10-12 at 10 26 05 pm

eirslett:
:+1: Good work!

adriancole:
ok gonna cut 0.4.0 which is going to be great, as it will have redis, high-precision timestamps and the dep graph fix!

adriancole:
shipped in v0.4.0

PR #55, opened 2016-11-13
No work in progress?

eirslett:
Looking at the development activity and usage statistics, I believe we can consider zipkin-js somewhat stable by now.


adriancole:
yeap!

PR #56, opened 2016-11-14
Upgrade to kafka-node@1.0.0

SimenB:
Breaking change is dropped support for node < 4, which is no problem here.

https://github.com/SOHU-Co/kafka-node/blob/master/CHANGELOG.md

/cc @sveisvei @eirslett


eirslett:
Sounds like a good idea to me. No API changes necessary?

SimenB:
Nope, shouldn't. Breaking upstream is just Node@4, and your tests should pick up other stuff, right?

eirslett:
:shipit:

PR #57, opened 2016-11-21
Propagates, reports and optionally generates 128-bit trace IDs

adriancole:
Before, we threw out high bits when receiving 128-bit trace IDs. This
retains and reports them to Zipkin.

When a Tracer is initialized with `traceId128Bit: true`, the trace IDs
will be 128-bit (32 characters) as opposed to 64-bit. In either case,
span IDs within a trace are 64-bit (16 characters).

To simplify things, this does away with variable length ID generation
in favor of the de-facto fixed length. It is unlikely that random
numbers would have a lot of leading zeros to optimize out anyway.

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -1,4 +1,4 @@
-const {None, fromNullable} = require('../option');
+const {None, Some, fromNullable} = require('../option');
 const {Sampler, alwaysSample} = require('./sampler');
 
 const Annotation = require('../annotation');
@@ -15,10 +15,16 @@ class Tracer {
   constructor({
     ctxImpl = requiredArg('ctxImpl'),
     recorder = requiredArg('recorder'),
-    sampler = new Sampler(alwaysSample)
+    sampler = new Sampler(alwaysSample),
+    // traceID128Bit enables the generation of 128 bit traceIDs in case the tracer
+    // needs to create a root span. By default regular 64 bit traceIDs are used.
+    // Regardless of this setting, the library will propagate and support both
+    // 64 and 128 bit incoming traces from upstream sources.
+    traceId128Bit = false
   }) {
     this.recorder = recorder;
     this.sampler = sampler;
+    this.traceId128Bit = traceId128Bit;
     this._ctxImpl = ctxImpl;
     this._defaultTraceId = this.createRootId();
     this._startTimestamp = now();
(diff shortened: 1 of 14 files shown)


packages/zipkin/src/zipkinCore.thrift
```diff
@@ -1,46 +1,448 @@
 # Generate with the command "thrift -gen js:node zipkinCore.thrift"
 namespace java com.twitter.zipkin.gen
-namespace rb Zipkin
 
-#************** Collection related structs **************
-
-# these are the annotations we always expect to find in a span
+#************** Annotation.value **************
```
adriancole: all this here is the result of bumping to the latest thrift (which has comments and a few new things)

adriancole:
@basvanbeek ptal

packages/zipkin/src/tracer/index.js
```diff
@@ -15,10 +15,16 @@ class Tracer {
   constructor({
     ctxImpl = requiredArg('ctxImpl'),
     recorder = requiredArg('recorder'),
-    sampler = new Sampler(alwaysSample)
+    sampler = new Sampler(alwaysSample),
+    // traceID128Bit enables the generation of 128 bit traceIDs in case the tracer
+    // needs to create a root span. By default regular 64 bit traceIDs are used.
+    // Regardless of this setting, the library will propagate and support both
+    // 64 and 128 bit incoming traces from upstream sources.
```
adriancole: description copied from https://github.com/openzipkin/zipkin-go-opentracing

eirslett:
Lgtm!

packages/zipkin/test/internalRepresentations.test.js
```diff
@@ -176,3 +179,49 @@ describe('JSON Formatting', () => {
...
+    expect(spanJson.traceId).to.equal(serverSpan.traceId.traceId);
+  });
+
+  it('should write trace ID as high and low numbers in Thrift representation', () => {
+    const serialized = serializeSpan(serverSpan, format = 'raw');
+
+    const spanThrift = new thriftTypes.Span();
+    spanThrift.read(new TBinaryProtocol(new TFramedTransport(serialized)));
+
+    expect(spanThrift.trace_id_high.toNumber(true)).to.equal(1234);
```
sveisvei: Yeah, this is annoying with thrift :(. node-int64 does not work as a number even if they claim to do so.

basvanbeek: welcome to the lovely world of Javascript where all numbers are float64

adriancole:
ok in it goes. thanks for having a look compadres

PR #59, opened 2016-11-25
BathRecorder: Do hostname IPv4 lookup before annotating span with Endpoint

### LINKED ISSUE #58: BatchRecorder expects ServerAddr annotation to be InetAddress
`record` method of BatchRecorder has [a piece of code][1], that tries to cast the `host` field of the annotation to integer, using `rec.annotation.host.toInt()` of your `InetAddress` class. 

It seems `ServerAddr` annotation constructor doesn't have any opinion about `host` parameter, so this behaviour mightn't be expected. 

For example, the `zipkin-instrumentation-redis` provides an example, which uses host as a raw string both for connection options and for remote server annotation:

```
const redisConnectionOptions = {
  host: 'localhost',
  port: '6379'
};
const redis = zipkinClient(tracer, Redis, redisConnectionOptions);
```

Shouldn't `BatchRecorder` cast host field of `ServerAddr` annotation to integer (ipv4/ipv6 address) during the recording process?

[1]: https://github.com/openzipkin/zipkin-js/blob/d446c82b962c33a98f134585d29c590b3707119f/packages/zipkin/src/batch-recorder.js#L110-L115

narqo:
This PR introduces an ability to use hostnames in `LocalAddr` and `ServerAddr` annotations instead of `InetAddr` instanses.

Since now on `BathRecorder` uses `Endpoint.createEndpoint()` factory, which do async IPv4 lookup (in case hostname was passed as `host`), before creating an `Endpoint` instance. While API of `BathRecorder` stayed the same, it's `record()` method is doing MutableSpan uptation asyncronosly, now.

closes #58

### DIFF
--- a/packages/zipkin/src/internalRepresentations.js
+++ b/packages/zipkin/src/internalRepresentations.js
@@ -1,8 +1,9 @@
 const thriftTypes = require('./gen-nodejs/zipkinCore_types');
 const {now, hrtime} = require('./time');
 const {Some, None} = require('./option');
+const {getAddressByName} = require('./InetAddress');
 
-function Endpoint({serviceName, host, port}) {
+function Endpoint({serviceName, host, port} = {}) {
   this.serviceName = serviceName;
   this.host = host || 0;
   this.port = port || 0;
@@ -27,6 +28,17 @@ Endpoint.prototype.toJSON = function toJSON() {
     ipv4: formatIPv4(this.host),
     port: this.port
   };
+};
+
+Endpoint.createEndpoint = function createEndpoint({serviceName, host, port}, callback) {
+  getAddressByName(host, addr => {
+    const ep = new Endpoint({
+      serviceName,
+      host: addr ? addr.toInt() : addr,
+      port,
+    });
+    callback(ep);
+  });
 };
 
 function ZipkinAnnotation({timestamp, value, endpoint}) {
@@ -116,7 +128,7 @@ function MutableSpan(traceId) {
   this.startTick = hrtime();
   this.name = None;
   this.service = None;
-  this.endpoint = new Endpoint({});
+  this.endpoint = new Endpoint();
   this.serverAddr = None;
   this.annotations = [];
   this.binaryAnnotations = [];
--- a/packages/zipkin/test/InetAddress.test.js
+++ b/packages/zipkin/test/InetAddress.test.js
@@ -13,4 +13,26 @@ describe('InetAddress', () => {
     const addr = new InetAddress('80.91.37.133');
     expect(addr.toString()).to.equal('InetAddress(80.91.37.133)');
   });
+
+  it('should get the address by hostname', done => {
+    InetAddress.getAddressByName('localhost', addr => {
+      expect(addr).to.be.an.instanceof(InetAddress);
+      done();
+    });
+  });
+
+  it('should get the address by IPv4 address', done => {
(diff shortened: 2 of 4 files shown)


packages/zipkin/src/internalRepresentations.js
```diff
@@ -1,8 +1,9 @@
 const thriftTypes = require('./gen-nodejs/zipkinCore_types');
 const {now, hrtime} = require('./time');
 const {Some, None} = require('./option');
+const {getAddressByName} = require('./InetAddress');
 
-function Endpoint({serviceName, host, port}) {
+function Endpoint({serviceName, host, port} = {}) {
```
adriancole: maybe make a code comment about what this is for?

packages/zipkin/src/internalRepresentations.js
```diff
@@ -29,6 +30,17 @@ Endpoint.prototype.toJSON = function toJSON() {
   };
 };
 
+Endpoint.createEndpoint = function createEndpoint({serviceName, host, port}, callback) {
+  getAddressByName(host, addr => {
```
adriancole: I'd either leave the field as an ip addr (converting lazily) or rename the function to getIpV4. Remember we do actually support ipv6 and some won't have a ipv4 address, but this code seems to help in either case.

ps it seems like this should just affect the host field, not the entire endpoint.

packages/zipkin/test/InetAddress.test.js
```diff
@@ -13,4 +13,26 @@ describe('InetAddress', () => {
     const addr = new InetAddress('80.91.37.133');
     expect(addr.toString()).to.equal('InetAddress(80.91.37.133)');
   });
+
+  it('should get the address by hostname', done => {
```
adriancole: maybe add a test where the lookup fails, but doesn't blowup the caller?

packages/zipkin/src/internalRepresentations.js
narqo: This was added to prevent the need of doing `new Endpoint({})` below, but I'm not happy with both variants. Will revert it, for now.

packages/zipkin/src/internalRepresentations.js
narqo: > ps it seems like this should just affect the host field, not the entire endpoint.

Could you describe this in more details, I'm not sure I've caught the point. 

We have to have a place where `host` will be set correctly. We can do this in `BachRecorder` and leave `internalRepresentations` completely free from async host-to-ip knowledge. 

Worth saying, I don't have any strong opinion about this part, so suggestions are totally welcome.

narqo:
> not sure about the impact of the flow change in batch-recorder

After spending a little bit more time thinking about the changes, I'm not sure if they are safe. It feels that additional checks in BatchRecorder are needed, to prevent duplicate span logging if IP address lookup process takes more time than it's allowed by the `timeout`.

packages/zipkin/src/internalRepresentations.js
adriancole: what I mean is that the host (or actually what's derived from it which is either an ipv4 or ipv6 address) is lower value data. The service name is the most important thing, so it would be nice to contain the fanciness to the host. In other words, a failure to parse what we are calling "host" shouldn't make endpoint fail to be set.

adriancole:
so.. are we looking for another option at this point?

adriancole:
closing until interest resumes

PR #62, opened 2016-12-12
Fix incorrect span durations when using hapi.js middleware

### LINKED ISSUE #61: Span durations appear to be incorrect for hapi.js middleware
Lengthy requests exceeding several seconds are being reported as 1-2ms. 

Wrote a failing test here: https://github.com/zackangelo/zipkin-js/blob/fix_hapi_span_duration/packages/zipkin-instrumentation-hapi/test/integrationTest.js#L157

Has anyone else noticed this? If so, or seems entirely possible, will open a PR to fix.

zackangelo:
Fixes #61. 

Not sure if what I'm doing here is okay. In order to split the span actions into two callbacks, I'm storing the trace on the request object and then reattaching it to the context in the response hook. 

@adriancole @eirslett

### DIFF
--- a/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
+++ b/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
@@ -34,9 +34,10 @@ exports.register = (server, {tracer, serviceName = 'unknown', port = 0}, next) =
     return;
   }
 
-  server.ext('onPreResponse', (request, reply) => {
-    const {headers, response} = request;
+  server.ext('onRequest', (request, reply) => {
+    const {headers} = request;
     const readHeader = headerOption.bind(null, headers);
+    const plugins = request.plugins;
 
     tracer.scoped(() => {
       if (readHeader(Header.TraceId) !== None && readHeader(Header.SpanId) !== None) {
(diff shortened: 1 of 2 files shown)


zackangelo:
Fixing linter errors.

zackangelo:
Seems that using the `.plugins` hash on the request object is the correct way for a middleware to inject request state onto the object.

> plugins - plugin-specific state. Provides a place to store and pass request-level plugin data. The plugins is an object where each key is a plugin name and the value is the state.

sveisvei:
Looks good to me. 

Indeed @zackangelo , seems that `request.plugins` is the convention for appending data from plugins.

zackangelo:
:warning: Please don't merge yet, may have found an issue.

zackangelo:
Moved the `reply.continue` call into `tracer.scoped` to keep the continuation chain in tact. Also, some versions of hapi were complaining about the `response.on('finish'...)` event hook so I'm just calling it in `onPreResponse` directly now (which seems to match other instrumentations more closely as well).

adriancole:
so good to go?

adriancole:
thanks @zackangelo

PR #63, opened 2016-12-13
Modify generated thrift files to fix global object leak

rooftopsparrow:
Because of [a bug in the thrift compiler](https://issues.apache.org/jira/browse/THRIFT-1840), it generates javascript that leaks objects into the global context. This patch simply edits the generated thrift files to fix that ( hopefully all of them ).


adriancole:
thx!

rooftopsparrow:
thank you! 🎉

Issue #16, opened 2016-06-17
Open to an ES5 transpile build step


guyellis:
If I (or someone) can work out a way to do it would you be open to accepting a PR that has a prepublish step that transpiles the ES6 code to ES5 so that ES5 restrained consumers can also use this module?

E.g. prepublish uses babel to generate output to /lib/es5/ before publish and that's the entry point for ES5 consumers. This would not impact the ES6 consumers and the current ES6 being used in the module. i.e. ES6 by default. ES5 needs to be explicitly `required` by the consumer by path.

Might also help with #1 which might need ES5 for browsers.

sveisvei:
Shipping a es5 version should be fine, and is probably needed for a universal version. I am all for it atleast

jcchavezs:
Any input @DavideCarvalho @sonybinhle?

DavideCarvalho:
I started a branch for this task. I think the best tech for this is rollup, since we can transpile for es5 and es6 at the same time. If someone needs to target only evergreen targets, they can use es6 build (and use stuff like type=module for scripts), otherwise, use es5.

sonybinhle:
I am for rollup also.

jcchavezs:
Awesome. I will link this issue in the rollup one. Thanks guys!

Den fre. 23. nov. 2018, 11:20 skrev Binh Le <notifications@github.com:

PR #64, opened 2016-12-19
Prepublish script for es5

tonypizzicato:
See #16


package.json
```diff
@@ -14,6 +14,8 @@
   "author": "OpenZipkin <openzipkin.alt@gmail.com>",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "devDependencies": {
+    "babel-cli": "^6.18.0",
+    "babel-preset-es2015": "^6.18.0",
```
eirslett: I think we should be able to get away with only using babel-preset-env and then target node 4.

eirslett: Another option would be to just run the whole code base through babel to make it node 4-compatible, and then just set up the build process etc. to use node 4. So we don't need a build step.

tonypizzicato: ok, makes sense. 
but pr was opened for browser specific packages support and running transpiling step before publishing to npm

eirslett: Yes, using it from browsers is a good point... in that case it's probably best to do it like you have, use babel-preset-es2015 and let that be used from both browsers and node servers.

tonypizzicato: so, what do you think about merging this pr and publishing new version to npm?

eirslett: LGTM! @adriancole?

adriancole:
maybe add a description to the PR and fix the type in the commit message? :) I added the issue link.

testing locally now. thanks for the help!

adriancole:
also cc @openzipkin/devops-tooling

adriancole:
so nothing broke for me, though my ability to test this is limited. Does this need any README changes, yet? Ex the readme talks about node 6

eirslett:
Looks like the build is failing because eslint is linting the generated code. We only want to lint the original code.

sveisvei:
Btw, this PR should use `git mv` so we don`t loose history.

tonypizzicato:
@sveisvei update pr with two commits: one with `git mv` that is treated as file renaming. but after changes in the second pr(fixing `require` paths) it's treaded as added/deleted files again compared to master

adriancole:
thanks for the update. ps lint config still needs an update as the build is broke

adriancole:
this change is still pending lint

tonypizzicato:
@adriancole updated .eslintignore

adriancole:
thanks for the work and follow-up @tonypizzicato

PR #66, opened 2016-12-26
Use isomorphic-fetch instead node-fetch

tonypizzicato:
Use isomorphic-fetch instead node-fetch for better browser support

### DIFF
--- a/packages/zipkin-instrumentation-express/package.json
+++ b/packages/zipkin-instrumentation-express/package.json
@@ -11,7 +11,7 @@
   "repository": "https://github.com/openzipkin/zipkin-js",
   "devDependencies": {
     "express": "^4.13.4",
-    "node-fetch": "^1.5.1",
+    "isomorphic-fetch": "^2.2.1",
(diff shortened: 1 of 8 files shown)


eirslett:
Sounds like a good idea.
Is it possible to test this in a browser? Maybe we should add karma configuration to the project, and run browser tests in PhantomJS?

adriancole:
>
> Is it possible to test this in a browser? Maybe we should add karma
> configuration to the project, and run browser tests in PhantomJS?
>
agreed. while extremely tempting to not add tests or punt to a future PR,
this is how tests don't end up backfilled despite best intentions. If you
can, please be a zipkin-js hero and try to test this!

adriancole:
ping back if this is a help-wanted type of status. Maybe someone can point to a similar project which is setup to do browser testing with PhantomJS

adriancole:
soo.. should we merge w/o test or ...

adriancole:
closing but will happily re-open when interest resumes

PR #68, opened 2016-12-31
Added http.req_id annotation to restify middleware

m64253:
This would allow us to much easier find the trace in UI by just searching for `http.req_id=…`.

### DIFF
--- a/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
+++ b/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
@@ -70,6 +70,7 @@ module.exports = function restifyMiddleware({tracer, serviceName = 'unknown', po
 
       tracer.recordServiceName(serviceName);
       tracer.recordRpc(req.method);
+      tracer.recordBinary('http.req_id', req.getId());
       tracer.recordBinary('http.url', url.format({
         protocol: req.isSecure() ? 'https' : 'http',
         host: req.header('host'),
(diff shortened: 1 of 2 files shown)


eirslett:
Where does req_id come from? Is it something restify-specific that the framework creates/injects?
Is it basically the same as a Zipkin span ID? Except somehow easier to extract?

m64253:
>Where does req_id come from? Is it something restify-specific that the framework creates/injects?
Is it basically the same as a Zipkin span ID? Except somehow easier to extract?

Yes almost the same as a Zipkin span ID, with the only difference that req_id is included the default request logger in restify. Which means that you can from logs get to a specific trace more easily.

eirslett:
Ok, I see. Is it also an UUID, like the Zipkin trace ID? Is there any way to set up restify so that it uses Zipkin's trace UUID as request UUID? Then it would be even simpler, because there would be only one ID instead of two?

m64253:
Yes that would preferable but I do not think it's currently possible to configure restify like that, at least not in v4 😞

m64253:
Found a solution that works for now, closing issue.

eirslett:
What was the solution? Maybe it's helpful to others?

m64253:
Yes sure, all I did was wrap the next callback and call `tracer.recordBinary(…)` inside of that.

```javascript
const zipkinMiddleware = restifyMiddleware({
  tracer,
  serviceName: 'My Service'
});

server.use(function zipkin(req, res, next) {
  zipkinMiddleware(req, res, () => {
    tracer.recordBinary('http.req_id', req.id());
    next();
  });
});
```

eirslett:
Ah, I see. Looks like a good solution. So recording req_is is opt-in/a plugin/middleware.

adriancole:
maybe add to the README?

PR #69, opened 2017-01-02
Listen for both close or finish

m64253:
See this: https://github.com/nodejs/node/issues/6929

### DIFF
--- a/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
+++ b/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
@@ -85,14 +85,14 @@ module.exports = function restifyMiddleware({tracer, serviceName = 'unknown', po
       const onCloseOrFinish = () => {
         res.removeListener('close', onCloseOrFinish);
         res.removeListener('finish', onCloseOrFinish);
-        
+
         tracer.scoped(() => {
           tracer.setId(id);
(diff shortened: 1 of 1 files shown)


packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
```diff
@@ -82,13 +82,19 @@ module.exports = function restifyMiddleware({tracer, serviceName = 'unknown', po
         tracer.recordBinary(Header.Flags, id.flags.toString());
       }
 
-      res.on('finish', () => {
+      const onCloseOrFinish = () => {
+        res.removeListener('close', onCloseOrFinish);
```
eirslett: Do we need to remove the listeners? My understanding is that `res.once(...)` will remove them for you, after invoking them?

m64253: Yes but the issue is that either `close` or `finish` get triggered so I opted remove both just to be safe.

m64253:
Need help resolving the appvayor erros 😞

eirslett:
Don't worry about the AppVeyor build, it has been broken for some time now. It's due to memcached not being installed on the CI server, which is unrelated to this PR.

adriancole:
thanks for the help!

PR #70, opened 2017-01-02
Try to fix Windows build

eirslett:
This is just me trying to fix the Windows build. It's not really a PR, I'll just keep pushing commits here (to trigger AppVeyor builds) until I can make it work.


eirslett:
Ok then, build is green!
The problem was that `choco install memcached` has ceased to work on AppVeyor. So instead, we start memcached manually if we're on windows. (So on non-windows based systems we assume that memcached is running, which is not necessarily true on developer machines, but at least on Travis it is)
I commited the memcached binaries directly into the repo; they were not too big (300 kb or so).

@adriancole review?

adriancole:
sure thing. will take a look tomorrow. thanks, Eirick!

adriancole:
I love it. pragmatic and easy to restore if choco starts working again

PR #77, opened 2017-02-28
fix sample example readme error

huanli2:


adriancole:
@huanli2 thanks so much!

PR #79, opened 2017-03-03
Add express-http-proxy instrumentation support

muymoo:
Standardize express instrumentation request recording


README.md
```diff
@@ -74,6 +74,8 @@ Java installed on your machine.
 
 Running code style linting: `npm run lint`
 
+AppVeyor is currently broken and ignored. PR welcome from those with Windows boxes.
```
adriancole: heh :P

packages/zipkin-instrumentation-express/README.md
```diff
@@ -22,3 +22,55 @@ app.use(zipkinMiddleware({
...
+
+const ctxImpl = new CLSContext();
+const recorder = new ConsoleRecorder();
+const tracer = new Tracer({ctxImpl, recorder});
+const nameOfApp = 'weather-app';
+const nameOfRemoteService = 'weather-api';
+const proxyInstrumentation = new ProxyInstrumentation(tracer, nameOfApp, nameOfRemoteService);
+
+app.use('/api/weather',
+  zipkinMiddleware({tracer, nameOfApp}),
```
adriancole: seems like we are passing the same arguments twice.. would it make sense to be able to derive the proxy instrumentation from the middleware?

packages/zipkin-instrumentation-express/src/expressMiddleware.js
```diff
@@ -79,7 +79,7 @@ module.exports = function expressMiddleware({tracer, serviceName = 'unknown', po
       const id = tracer.id;
 
       tracer.recordServiceName(serviceName);
-      tracer.recordRpc(req.method);
+      tracer.recordRpc(req.method.toUpperCase());
```
adriancole: this won't really help as zipkin will downcase it!

packages/zipkin-instrumentation-express/README.md
muymoo: How are you thinking you would instantiate it? Possibly like:
```javascript
const expressInstrumentation = new ExpressInstrumentation({tracer, nameOfApp, nameOfRemoteService});
const zipkinMiddleware = expressInstrumentation.middleware;
const proxyInstrumentation = new expressInstrumentation.ProxyInstrumentation();
```
Do you think all users of the proxy will use the middleware too?

adriancole:
>
> const expressInstrumentation = new ExpressInstrumentation({tracer, nameOfApp, nameOfRemoteService});const zipkinMiddleware = expressInstrumentation.middleware;const proxyInstrumentation = new expressInstrumentation.ProxyInstrumentation();
>
> yep something like this, though unclear why the latter needs to be a new
object vs a constant

> Do you think all users of the proxy will use the middleware too?
>
 no idea actually. I don't know how people use express in general, so if
what I'm mentioning isn't sensible in practice, nevermind!

packages/zipkin-instrumentation-express/src/expressMiddleware.js
muymoo: All the other zipkin-js libraries are doing it? Should we remove this across them? 
Examples:
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-fetch/src/wrapFetch.js#L27

https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js#L37

adriancole: yeap

adriancole:
I've not heard others explicitly mentioning anything on this PR. I like what you've done and also the README. I'll merge and wait a day to release (in case someone wakes up surprised!)

DailunCheng:
@muymoo 
Hi could you explain a little bit more about your last line in readme for me? 
"app.use('/api/weather', expressMiddleware({tracer, serviceName}), zipkinProxy('http://api.weather.com'));"
1. What is in '/api/weather'? 
2. Would the proxy works for local inner network endpoints? Ex: I am having a docker-compose network with a nodejs front-end container doing proxy for multiple backend containers, I suppose the URL argument should be the endpoint URL of one container. Am I understanding correctly for this http proxy use case?

muymoo:
@DailunCheng 
1. `/api/weather` is an example route this express server exposes. The instrumented `expressMiddleware` adds tracing info to the incoming request. The `zipkinProxy` is a wrapped instance of `express-http-proxy` that will proxy the request to `http://api.weather.com`
2. Yes. I imagine that would work although I haven't tried the docker use case specifically. The argument of `zipkinProxy` would be the backend container URL you'd like to hit.

DailunCheng:
@muymoo
Thanks for clarification! I tried it but seems like one nodejs function the instrumentation used has been deleted and the browser is complaining about it :( I try to work around it by using zipkin-instrumentation-cujojs-rest and it works for me! However I am still looking forward to use express native way to solve my problem!

PR #80, opened 2017-03-05
Fix crash when hapi responds with error

rexxars:
In hapi, you can respond with an error, which will be wrapped by [Boom](https://github.com/hapijs/boom). In this case, `response.statusCode` is undefined, since `response` is an instance of a Boom error.

Currently, it crashes when going to a route that produces an error (such as a 404) due to calling `toString()` on undefined. This PR fixes this by checking if the response is a Boom instance and retrieving the status code from that.

### DIFF
--- a/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
+++ b/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
@@ -93,10 +93,11 @@ exports.register = (server, {tracer, serviceName = 'unknown', port = 0}, next) =
 
   server.ext('onPreResponse', (request, reply) => {
     const {response} = request;
+    const statusCode = response.isBoom ? response.output.statusCode : response.statusCode;
 
     tracer.scoped(() => {
       tracer.setId(request.plugins.zipkin.traceId);
-      tracer.recordBinary('http.status_code', response.statusCode.toString());
+      tracer.recordBinary('http.status_code', statusCode.toString());
       tracer.recordAnnotation(new Annotation.ServerSend());
     });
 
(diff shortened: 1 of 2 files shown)


packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
```diff
@@ -93,10 +93,11 @@ exports.register = (server, {tracer, serviceName = 'unknown', port = 0}, next) =
 
   server.ext('onPreResponse', (request, reply) => {
     const {response} = request;
+    const statusCode = response.isBoom ? response.output.statusCode : response.statusCode;
```
adriancole: note.. if boom is an error, you could add a binary annotation "error" (Constants.ERROR) with the message

rexxars: Interesting, is this usually done for all errors (client errors as well as server errors)?

rexxars:
As far as I know, there isn't really any default handling of unhandled exceptions in hapi. There are many different plugins that can be installed that does everything from heap dumps to error formatting etc, so it's hard to make something that works for every case.

Hapi does have built-in support for [responding to requests with an error](https://hapijs.com/api#error-response), which my PR handles.

I'd be interested in knowing more about the error annotation you mentioned - doesn't look like the express instrumentation handles errors, do you have any examples of how this works? When is the annotation commonly used? For both server and client errors, or just server errors?

adriancole:
>
> I'd be interested in knowing more about the error annotation you mentioned
> - doesn't look like the express instrumentation handles errors, do you have
> any examples of how this works? When is the annotation commonly used? For
> both server and client errors, or just server errors?
>
The important part is that "error" colors the span red in the UI, which
makes it easy to see there's an error there. The value of "error" is less
important (some write "server error" while others format a message). This
is slightly more important for client spans as they represent the critical
path, but there's no harm in adding it twice.

Here's an example of another recent change doing this..
https://github.com/hyleung/ratpack-zipkin/pull/18

I agree that there's no existing code here in zipkin-js, but help wanted!

PR #81, opened 2017-03-08
Add —force workaround for AppVeyor

muymoo:
Trying ideas from https://github.com/npm/npm/issues/10826


adriancole:
thanks! will merge in 24hrs unless someone screams.

rexxars:
Don't think this is necessary anymore, since the windows build seems to be running now after #82 was merged.

PR #82, opened 2017-03-08
Hoist dev dependencies with lerna

rexxars:
This PR turned out a little bigger than I intended it to be, however:

* Install mocha and babel as dev dependencies for each project, then use lerna's hoist feature to simply symlink the dependencies down to the child modules. This enables us to run tests and babel within each module without using parent pathing, which doesn't work well on windows (../../node_modules etc)
* Simplified scripts - no need to specify full path to dependency binaries, as npm puts the `.bin` folder in the path when executing
* The build task that was attempting to run babel didn't specify any presets, so nothing actually got compiled to ES5 as far as I could tell. Added a babel config to make this work as expected
* Made a few modules use the same versions for dependencies (mocha, node-fetch) in order to hoist dependencies more efficiently

This *might* fix the long broken windows build, and makes things slightly more "clean", in my opinion.


rexxars:
@muymoo The tests for express http proxy seems to be failing on node 7.
Not sure what the deal is, here's some info:

```
AssertionError: expected '127.0.0.1:59291/weather?index=10&count=300' to equal '//127.0.0.1:59291/weather?index=10&count=300'
+ expected - actual

-127.0.0.1:59291/weather?index=10&count=300
+//127.0.0.1:59291/weather?index=10&count=300
```

rexxars:
@eirslett Any suggestions on what might be the deal with the kafka tests [on windows](https://ci.appveyor.com/project/eirslett/zipkin-js/build/1.0.218)? I find it very hard to read the logs/code, and it works on travis, so...

muymoo:
@eirslett It seems that node url changed between 6 and 7. Here's the important lines in url.js:

v6
```javascript
  if (this.slashes ||
      (!protocol || slashedProtocol[protocol]) && host !== false) {
    host = '//' + (host || '');
    if (pathname && pathname.charCodeAt(0) !== 47/*/*/)
      pathname = '/' + pathname;
  } else if (!host) {
    host = '';
  }
```

v7
```javascript
  if (this.slashes || slashedProtocol[protocol]) {
    if (this.slashes || host) {
      if (pathname && pathname.charCodeAt(0) !== 47/*/*/)
        pathname = '/' + pathname;
      host = '//' + host;
    } else if (protocol.length >= 4 &&
               protocol.charCodeAt(0) === 102/*f*/ &&
               protocol.charCodeAt(1) === 105/*i*/ &&
               protocol.charCodeAt(2) === 108/*l*/ &&
               protocol.charCodeAt(3) === 101/*e*/) {
      host = '//';
    }
  }
```
The difference in v7 is that it will not add `//` if the protocol doesn't exist. In v6, `//` was added even if the protocol didn't exist but there was a host name `(!protocol || ...)`.

Options to resolve:
1. Add `slashes:true` to https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-express/src/wrapExpressHttpProxy.js#L7. This should not be a problem assuming the proxy will always be used for http(s). Express-http-proxy doesn't allow us to get at the client request protocol.
2. File bug with node v7 for regression here: https://github.com/nodejs/node/commit/336b027411b2f413e02033720bbc4a5f3f0d96bd#diff-bbc5176adff1bbc01acd61bfc85d049fL606

rexxars:
@adriancole I am considering this "ready for merge" now, as the tests aren't running at all in master, but with the changes outlined here they actually run, but fail. That gives us a much easier time trying to figure out *why* it's failing.

On another note, would it be an idea to deprecate the scribe transport? As I understand it, scribe is kind of on the way out. [zipkin-transport-scribe])(https://www.npmjs.com/package/zipkin-transport-scribe) has ~70 downloads a month, the node library we're using behind the scenes have not been published in a new version for 4 years, and it's generating weird errors on the windows build that I can't seem to figure out.

I'm thinking it might cost us more to try to maintain than what it gains us.

adriancole:
Scribe is indeed a historic transport and i am unsure about how much usage
it gets in js. It might be well enough to introduce some "skip"
functionality which would be commented as ok as already tested in linux

adriancole:
looks like an improvement to me. does any of this invalidate or change instructions in the README?

rexxars:
Nope, still the same steps.

PR #83, opened 2017-03-09
Slashes: true to resolve url format breaking changes between node v6/7

muymoo:
Node url changed its formatting between v6 and 7. See https://github.com/nodejs/node/issues/11103

This will maintain backwards compatibility with 6 and 7. @rexxars


adriancole:
seems legit to me.. can someone else ack?

rexxars:
Looks good to me!

PR #84, opened 2017-03-09
Make tests runnable on node 4

rexxars:


sveisvei:
Wfm, but node 4 support is on the last days.

rexxars:
LTS for node 4 ends in April, yes, but it'll still be in maintenance up until April 2018, so I assume there will be people still using it, and I think we should be able to support them.

PR #86, opened 2017-03-22
Feature/request instrumentation

muymoo:
Adds instrumentation for https://github.com/request/request

Appreciate feedback. This is a preliminary attempt at instrumenting request. I was struggling to figure out a good way to clean it up and share the trace id between the request and response. Seems to be a lot of duplicate code between client instrumentations.

### DIFF
--- a/packages/zipkin-instrumentation-request/package.json
+++ b/packages/zipkin-instrumentation-request/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "zipkin-instrumentation-express",
+  "name": "zipkin-instrumentation-request",
   "version": "0.6.2",
-  "description": "Express middleware for instrumentation with Zipkin.js",
+  "description": "Instrumentation for request with Zipkin.js",
   "main": "lib/index.js",
   "scripts": {
     "build": "babel src -d lib",
--- a/packages/zipkin-instrumentation-request/package.json
+++ b/packages/zipkin-instrumentation-request/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "zipkin-instrumentation-express",
+  "version": "0.6.2",
+  "description": "Express middleware for instrumentation with Zipkin.js",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "babel src -d lib",
+    "test": "mocha --require ../../test/helper.js",
+    "prepublish": "npm run build"
+  },
+  "author": "OpenZipkin <openzipkin.alt@gmail.com>",
+  "license": "Apache-2.0",
+  "repository": "https://github.com/openzipkin/zipkin-js",
+  "devDependencies": {
+    "babel-cli": "^6.23.0",
+    "express": "^4.13.4",
+    "mocha": "^3.2.0",
+    "request": "^2.81.0",
+    "zipkin": "^0.6.1"
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/zipkin-instrumentation-request/src/wrapRequest.js
```diff
@@ -0,0 +1,34 @@
+const {Request, Annotation} = require('zipkin');
+
+function wrapRequest(request, {tracer, serviceName = 'unknown', remoteServiceName}) {
+  return request.defaults((options, callback) => {
+    tracer.scoped(() => {
+      tracer.setId(tracer.createChildId());
+      const traceId = tracer.id;
```
muymoo: What's the recommended way to share trace id across request/response? Does it have to fall within a single `tracer.scoped(...)`? I see some use `this.traceId` (ex. cujojs) and then others use `const traceId` (ex. fetch).

adriancole: when you are writing instrumentation, there are two concerns about propagating the traceId:
* ensuring async/error flow are attached to the correct id (manual propagation like you have is fine)
* ensuring downstream library calls are associated with the right id (scoped)

The latter isn't obvious until modification or child the client span is attempted by third-party code. Ex a client span could have a child for TLS setup, for example. If the current traceId wasn't scoped, such child would have an incorrect parent. I don't think we have tests like this at the moment.

adriancole: actually I'm wrong.. we do have tests for letContext :)

adriancole:
general next steps are to have others who use zipkin-js say they are interested in this. I'm almost positive there will be, but this is mainly to avoid single-user integration maintenance. Often, poking people is needed (people you know, people on gitter, etc)

travisby:
Highly desired to have request traced :).  I switched some of our stuff over to `fetch` since there was no `request` implementation at the time.  But we still have more and it'd be easier to use a wrapped request instead!

muymoo:
@travisby Glad to hear it is wanted. We have a bunch of apps using `request` too. If you get a chance, could you try it out and let me know how it works for you?

packages/zipkin-instrumentation-request/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "zipkin-instrumentation-request",
+  "version": "0.6.2",
```
adriancole: I don't know lerna that well.. when we go to publish, will it know this version hasn't been published yet? or should we add as the last version

packages/zipkin-instrumentation-request/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "prepublish": "npm run build"
+  },
+  "author": "OpenZipkin <openzipkin.alt@gmail.com>",
+  "license": "Apache-2.0",
+  "repository": "https://github.com/openzipkin/zipkin-js",
+  "devDependencies": {
+    "babel-cli": "^6.23.0",
+    "express": "^4.13.4",
+    "mocha": "^3.2.0",
+    "request": "^2.81.0",
```
adriancole: shouldn't this be in the dependencies section?

feifeiiiiiiiiiii:
LGTM

packages/zipkin-instrumentation-request/package.json
muymoo: The wrapper doesn't depend on `request`. The user is expected to pass in `request` as the first parameter to `wrapRequest()`. We only use depend on it for unit testing.

packages/zipkin-instrumentation-request/package.json
muymoo: I'm no lerna expert, but according to their docs, it will 

> Update the package.json of all updated packages to their new versions.
https://github.com/lerna/lerna#publish

I'm assuming this will publish `zipkin-instrumentation-request` as 0.6.2 since no version existed before.

```
zipkin-js $ lerna updated
Lerna v2.0.0-beta.38
Checking for updated packages...
Comparing with: v0.6.2

- zipkin-instrumentation-request
```

adriancole:
Cool thx for the info!

PR #89, opened 2017-03-24
Fixes not return right ctx

feifeiiiiiiiiiii:


eirslett:
? The code was correct the way it was.

PR #90, opened 2017-03-24
Fixes remove _session.run everything seemed right

feifeiiiiiiiiiii:


eirslett:
The reason the session.run stuff is there, is that if the callable() makes changes to the context, you know that the context value will be restored to what it was before running callable() after the scope is finished.
So if the value is A, and inside callable() you set the value to B, after scoped is finished, it will reset the value to be A.

PR #91, opened 2017-03-24
Use events instead of optional callback

muymoo:
White space aside, the biggest change is to use `.on('response', ...)` instead of wrapping the optional callback. Some libraries (like `express-request-proxy` which use `request` internally) do not always include a callback. There is the option to listen for a `response` event. This event is always emitted by `request` when a response is returned so I think it's a better approach then the original `wrappedCallback`.

### DIFF
--- a/packages/zipkin-instrumentation-request/src/wrapRequest.js
+++ b/packages/zipkin-instrumentation-request/src/wrapRequest.js
@@ -24,7 +24,15 @@ function wrapRequest(request, {tracer, serviceName = 'unknown', remoteServiceNam
       tracer.recordAnnotation(new Annotation.ClientRecv());
     };
 
-    return request(wrappedOptions, callback).on('response', recordResponse);
+    const recordError = (error) => {
+      tracer.setId(traceId);
+      tracer.recordBinary('request.error', error.toString());
+      tracer.recordAnnotation(new Annotation.ClientRecv());
+    };
+
+    return request(wrappedOptions, callback)
+      .on('response', recordResponse)
+      .on('error', recordError);
   }));
 }
 
--- a/packages/zipkin-instrumentation-request/src/wrapRequest.js
+++ b/packages/zipkin-instrumentation-request/src/wrapRequest.js
@@ -26,7 +26,7 @@ function wrapRequest(request, {tracer, serviceName = 'unknown', remoteServiceNam
 
     const recordError = (error) => {
       tracer.setId(traceId);
-      tracer.recordBinary('request.error', error.toString());
+      tracer.recordBinary('error', error.toString());
       tracer.recordAnnotation(new Annotation.ClientRecv());
     };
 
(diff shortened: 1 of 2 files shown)


packages/zipkin-instrumentation-request/src/wrapRequest.js
```diff
@@ -1,33 +1,31 @@
...
-      request(wrappedOptions, wrappedCallback);
-    });
-  });
+    const recordResponse = (response) => {
+      tracer.setId(traceId);
+      tracer.recordBinary('http.status_code', response.statusCode.toString());
+      tracer.recordAnnotation(new Annotation.ClientRecv());
+    };
+
+    return request(wrappedOptions, callback).on('response', recordResponse);
```
adriancole: what about on error?

muymoo:
I see "request.error" used elsewhere so I followed that convention although I don't see it anywhere in the thrift standards doc.

adriancole:
in the thrift, it is named "error"

packages/zipkin-instrumentation-request/src/wrapRequest.js
```diff
@@ -1,33 +1,39 @@
...
-  });
+    const recordResponse = (response) => {
+      tracer.setId(traceId);
+      tracer.recordBinary('http.status_code', response.statusCode.toString());
+      tracer.recordAnnotation(new Annotation.ClientRecv());
+    };
+
+    const recordError = (error) => {
+      tracer.setId(traceId);
+      tracer.recordBinary('request.error', error.toString());
```
adriancole: if you use the standard key "error" it makes the span red in the UI

adriancole: ps sometimes we run into things that aren't right in other integrations.. better to correct new ones than follow bad practice. even better to raise another PR to correct the others :)

packages/zipkin-instrumentation-request/test/integrationTest.js
```diff
@@ -147,4 +141,157 @@ describe('request instrumentation - integration test', () => {
...
+          expect(annotations[3].annotation.annotationType).to.equal('ClientSend');
+
+          expect(annotations[4].annotation.annotationType).to.equal('ServerAddr');
+
+          expect(annotations[5].annotation.annotationType).to.equal('BinaryAnnotation');
+          expect(annotations[5].annotation.key).to.equal('request.error');
+          expect(annotations[5].annotation.value)
+            .to.contain('Error: getaddrinfo ENOTFOUND bad.invalid.url bad.invalid.url:80');
+
+          expect(annotations[6].annotation.annotationType).to.equal('ClientRecv');
```
adriancole: maybe expect there is no annotations[7] (mainly paranoid the receive isn't added twice in error case)

muymoo:
RE: 'error' - I thought I should follow thrift doc but was also trying to be consistent :) Will update others to `error` in another PR.

adriancole:
well done!

PR #103, opened 2017-04-21
instrument http with zipkin-instrumentation-http

Dan70402:
This is a quick implementation using most the code from  zipkin-instrumentation-request


adriancole:
Thanks for raising this. needs votes from others who will use this before proceeding. feel free to nag people you see on github doing the same thing as here

adriancole:
you'll want to rebase as @pbadenski has some helpers for http clients now I think

adriancole:
this is over a year old. While probably good we should drop it until interest resumes. Thanks

Issue #107, opened 2017-07-07
Backfill LocalAddr when missing


adriancole:
Right now, only the server-side code add the annotation LocalAddr, usually when they have a port

```javascript
tracer.recordAnnotation(new Annotation.LocalAddr({port}));
```

https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-express/src/expressMiddleware.js#L85
 
The client side doesn't, but it should as otherwise annotations are allocated to IP "unknown"

One way is to go through all the client libraries and do the following when they don't know the client-side of the socket.
```javascript
tracer.recordAnnotation(new Annotation.LocalAddr());
```

Another way is to default LocalAddr somehow in tracer

Thanks to @bojand for reporting

jcchavezs:
@adriancole is this still valid based on the performance concerns from https://github.com/openzipkin/zipkin-js/pull/237?

PR #108, opened 2017-07-11
Add LocalAddr annotation and improve ServerAddr annotation for clients

bojand:
Hello, initial pass / review for changes to address #107 :
* Adds `LocalAddr` annotation for `request`, `cujojs-rest` and `fetch` instrumentations.
* Improves `ServerAddr` annotation for `request` module by listening for the socket connection event and then getting the actual [socket address](https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_socket_address) and storing it into the annotation. We also try to record this in case of an error using the request object's socket connection.
* In case that the socket is not available we just use the requests hostname. Not sure if this is favourable?
* Unfortunately the other 2 client libs (`cujojs-rest` and `fetch`) do not offer an api or any sensible way that exposes the raw socket at the time when the address would be available. `cujojs` exposes the raw request and response objects on the `response` intercept, but at that point the socket connection is closed and the socket does not have `address()` any more.
* Added tests for all of the above. All tests pass.
* Verified `request` usage manually in a [simple app](https://github.com/bojand/zipkin-express-sample) where I had discovered the initial problem.

### DIFF
--- a/packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js
+++ b/packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js
@@ -30,7 +30,7 @@ function request(req, {tracer, serviceName = 'unknown', remoteServiceName}) {
     tracer.recordBinary('http.url', reqWithHeaders.path);
     tracer.recordAnnotation(new Annotation.ClientSend());
     tracer.recordAnnotation(new Annotation.LocalAddr({port: URL.parse(reqWithHeaders.path).port}));
-    
+
     if (remoteServiceName) {
       // TODO: can we get the host and port of the http connection?
       tracer.recordAnnotation(new Annotation.ServerAddr({
--- a/packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js
+++ b/packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js
@@ -1,4 +1,5 @@
 /* eslint-disable no-param-reassign */
+const URL = require('url');
 const interceptor = require('rest/interceptor');
 const {
   Annotation,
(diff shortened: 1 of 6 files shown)


packages/zipkin-instrumentation-cujojs-rest/src/restInterceptor.js
```diff
@@ -28,6 +29,8 @@ function request(req, {tracer, serviceName = 'unknown', remoteServiceName}) {
     tracer.recordRpc(method.toUpperCase());
     tracer.recordBinary('http.url', reqWithHeaders.path);
     tracer.recordAnnotation(new Annotation.ClientSend());
+    tracer.recordAnnotation(new Annotation.LocalAddr({port: URL.parse(reqWithHeaders.path).port}));
```
adriancole: the localAddr port would be the client side of the socket (not the server side). So better to leave this out as it confuses things.

bojand:
Thanks for the feedback. Yup still trying to understand everything. Will try and improve and address things soon.

bojand:
I have updated the PR with the following changes:
* Use the socket remote address and remote port for Server Address annotation if we're recording it
* Use the connection socket local port in local address annotation

I hope I got this right as I am still not 100 % sure I completely understand what these annotations mean exactly. They do not seem to be documented [here](http://zipkin.io/pages/instrumenting.html)?

Can you please elaborate some more on what you mean exactly by setting the default local address in the tracer for spans when RPCs are made. I can try and make that improvement but I am afraid I'm still somewhat clueless about some details of the implementation.

adriancole:
The most detailed description of the annotations are here
https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift

I can have a look later tonight or tomorrow and reply more deeply. Thanks
for incrementing on this!

adriancole:
closing this as the code around it changed. have a look at model.js which maps to http://zipkin.io/zipkin-api/#/default/post_spans

PR #115, opened 2017-07-27
Remove unnecessary node core dependencies

### LINKED ISSUE #114: Compatability with non-node environments
Hey there,

First of all thanks for this great project, it really helps us tracing performance issues in our stack. For the last part of the stack, the react native mobile app, we ran into an issue:

[Zipkin requires](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/InetAddress.js#L1) the package `network-address` to determine the current IP automatically in [`getLocalAddress`](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/InetAddress.js#L23-L25). 

We don't need this functionality, but as the require is on top level we still run into issues, because `network-address` uses the `os` package. This is native to node and not available in the V8 / Safari engine we use.

If we could move the require into the function or migrate to a fallback for non-node environments it would be a lot easier for us. If you agree I would be glad to contribute a PR.

DanielMSchmidt:
Fixes #114 
@adriancole Could you take a look?

### DIFF
--- a/packages/zipkin/src/InetAddress.js
+++ b/packages/zipkin/src/InetAddress.js
@@ -1,5 +1,3 @@
-const networkAddress = require('network-address');
-
 class InetAddress {
   constructor(addr) {
     this.addr = addr;
@@ -20,7 +18,15 @@ class InetAddress {
   }
 }
 
+// In non-node environments we fallback to 127.0.0.1
 InetAddress.getLocalAddress = function getLocalAddress() {
+  const isNode = typeof process === 'object' && typeof process.on === 'function';
+  if (!isNode) {
+    return new InetAddress('127.0.0.1');
+  }
+
+  // eslint-disable-next-line global-require
+  const networkAddress = require('network-address');
   return new InetAddress(networkAddress.ipv4());
 };
 
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -1,5 +1,10 @@
 /* eslint-disable no-console */
-const fetch = require('node-fetch');
+const globalFetch =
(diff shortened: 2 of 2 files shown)


packages/zipkin/src/InetAddress.js
```diff
@@ -21,6 +19,8 @@ class InetAddress {
 }
 
 InetAddress.getLocalAddress = function getLocalAddress() {
+  // eslint-disable-line global-require
+  const networkAddress = require('network-address');
```
adriancole: what happens when this is called in a browser? :) is there an alternative or fallback approach we can use in browsers?

DanielMSchmidt: It would fail in the current implementation. It is only slightly better, but I honestly don't know of any solution for getting the network address in such a setting.

adriancole: maybe fallback to 127.0.0.1 instead?

DanielMSchmidt: Why not, will add that

adriancole:
thx also comment that this will fail in browsers (or some other way to say
the same thing, so that folks know why we are defensive).

afterwards go for gold!

Thanks for the help!

DanielMSchmidt:
@adriancole I think this looks fine now. Appveyor seems to be stuck, could you check? Would love to have a bugfix release out ;)

adriancole:
I'm .. gonna merge it :)

DanielMSchmidt:
Awesome, could you also do a release? 👍

adriancole:
0.7.3 on the way. thanks for the help!

DanielMSchmidt:
Thanks for supporting 👍

PR #117, opened 2017-07-28
Extract packages to reduce dependencies for browser based environments

DanielMSchmidt:
Hey there,

We wanted to get zipkin-js working in a React Native environment, therefore we needed to get rid of the node dependencies. 

I hope you like the way we chose to inject the dependencies into the modules, we tried to stay as close as possible to the existing implementation.

\cc @nicolai86

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -3,13 +3,11 @@ const globalFetch =
   (typeof window !== 'undefined' && window.fetch) ||
   (typeof global !== 'undefined' && global.fetch);
 
-// eslint-disable-next-line global-require
-const fetch = globalFetch || require('node-fetch');
-
 class HttpLogger {
-  constructor({endpoint, httpInterval = 1000}) {
+  constructor({endpoint, httpInterval = 1000, fetch = globalFetch}) {
     this.endpoint = endpoint;
     this.queue = [];
+    this.fetch = fetch;
 
     const timer = setInterval(() => {
       this.processQueue();
@@ -26,7 +24,7 @@ class HttpLogger {
   processQueue() {
     if (this.queue.length > 0) {
       const postBody = JSON.stringify(this.queue);
-      fetch(this.endpoint, {
+      this.fetch(this.endpoint, {
         method: 'POST',
         body: postBody,
         headers: {
--- a/packages/zipkin/src/InetAddress.js
+++ b/packages/zipkin/src/InetAddress.js
@@ -18,16 +18,4 @@ class InetAddress {
   }
 }
 
-// In non-node environments we fallback to 127.0.0.1
-InetAddress.getLocalAddress = function getLocalAddress() {
-  const isNode = typeof process === 'object' && typeof process.on === 'function';
-  if (!isNode) {
-    return new InetAddress('127.0.0.1');
-  }
-
-  // eslint-disable-next-line global-require
-  const networkAddress = require('network-address');
-  return new InetAddress(networkAddress.ipv4());
-};
-
 module.exports = InetAddress;
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -82,7 +82,7 @@ module.exports = function expressMiddleware({tracer, serviceName = 'unknown', po
       tracer.recordRpc(req.method.toUpperCase());
       tracer.recordBinary('http.url', formatRequestUrl(req));
       tracer.recordAnnotation(new Annotation.ServerRecv());
-      tracer.recordAnnotation(new Annotation.LocalAddr({port}));
+      tracer.recordLocalAddr({port});
 
       if (id.flags !== 0 && id.flags != null) {
         tracer.recordBinary(Header.Flags, id.flags.toString());
--- a/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
+++ b/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js
@@ -81,7 +81,7 @@ exports.register = (server, {tracer, serviceName = 'unknown', port = 0}, next) =
       tracer.recordRpc(request.method.toUpperCase());
       tracer.recordBinary('http.url', url.format(request.url));
       tracer.recordAnnotation(new Annotation.ServerRecv());
-      tracer.recordAnnotation(new Annotation.LocalAddr({port}));
+      tracer.recordLocalAddr({port});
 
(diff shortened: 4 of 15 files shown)


adriancole:
Thanks for helping with this. It seems implied there is something about batch recorder that doesnt gel? Can you enumerate what that is and what instead would be used in react (for batching spans). On one hand reducing deps is good on the other maybe there is a way to think of node as a platform vs splitting type by type (ex zipkin-platform-node?)

Maybe you can add a small example one of us can try, as a gist or repo? I'd like to be able to help maintain.

DanielMSchmidt:
The problem with the batch recorder is the dependencies. In a Browser / React Native environment we cannot use these node only dependencies.
We would need to come up with a separate solution for such an environment, but I think that would be very specific to the platform (e.g. using the React Native bridge to store it on the native side before sending it through or localstorage in the browser).

For us batching is not a problem, we would simply not batch it, but send the requests directly (we would only have one client span currently).

I could build a simple react native example, should I include it in this PR?

Thank you for your time <3

DanielMSchmidt:
@adriancole I think we had a small misconception, we thought we could hook up the logger directly to the tracer. We debugged through the dependencies and the main problem preventing us from simply using the batch recorder is that it uses thrift, which uses node-only things like util. 

As far as I understand it this means that we can't write our own recorder without copying the strings of thrift over and writing the serialization ourselves. Do you know of any "workaround" we could use for that problem? Otherwise we would need to go a completely other route :/

eirslett:
Have you tried using webpack or browserify? They replace node-specific dependencies with browser versions. Also, I think you can stub the dependencies so they're not imported. (I don't think they are always used either)

It might be possible to use something like react-native-thrift: https://www.npmjs.com/package/react-native-thrift

Or if possible, we could try to move the thrift dependency into the zipkin-transport-scribe package...

DanielMSchmidt:
@eirslett Do we only need thrift there?  
I haven't used browserify and I would love to not have to add an extra level of abstraction on my side.

eirslett:
We could probably get away with using browserify or webpack in the zipkin-js build, so we could publish builds without a webpack/browserify dependency.

DanielMSchmidt:
@eirslett Sounds like a good idea, but we didn't manage to get this working as some dependencies couldn't be inline to some reason. Can you help us with this?

adriancole:
>
> I could build a simple react native example, should I include it in this
> PR?
>
mind adding one as a different PR or as one here?
https://github.com/openzipkin/zipkin-js-example

this will help less acclimated like me validate things. right now ^^ uses
browserify and a very silly "frontend" having something more realistic
would be great.

DanielMSchmidt:
@adriancole will move this as soon as we have an idea how to move forward. I cannot take any more company time on this if I have no clear picture on how to get to a result :/

adriancole:
I think it is possible to move thrift to scribe (or a separate dep) and use
json instead as a base. It is also possible to write thrift directly
(binary by copying the java impl), but there is some effort there.

If thrift dep were gone, is that the main problem? I can help with things
like this even if I am less experienced with the overall landscape.

adriancole:
Ps appreciate your patience and also that you are trying this. I will help
to the best of my ability.

DanielMSchmidt:
Thank you so much @adriancole! I think with having thrift removed from the core module and the changes we have here (injecting fetch, e.g.) I think it would be possible to use zipkin.js in a React Native context. 

Regarding this PR, I would undo the commit that moves the thrift thing into a new package and I will move the demo into the other repository. Can you keep me up to date with the thrift stuff and if you need help there?

adriancole:
>
> Can you keep me up to date with the thrift stuff and if you need help
> there?
>
yep will do. thanks!

DanielMSchmidt:
Thank you so much for your time :)

DanielMSchmidt:
@adriancole I removed the batch recorder extraction and the example from this PR, does it look okay to you?

adriancole:
most looks good. only thing I'm concerned about is that for node (which is most users of this library), now they won't get the correct local address anymore. Is there no way to detect or patch in such a way that users won't have to change their code to get the correct IP?

side note: we should probably not list IP (ex use ipv4 0) vs 127.0.0.1 in hindsight

DanielMSchmidt:
@adriancole they would just need to add a new package, the functionality is still there. There is not as far as I can see. We can't resolve the dependencies of the local address package on compile time.

adriancole:
It seems like they need to patch the tracer, where they weren't before.

new zipkin.Tracer({
  localAddr: require('zipkin-local-address'),
})

This is different than a shim which you could do anywhere. Maybe
someone with more javascript experience than me can think of a way to
make it more automatic? I'd hate to make all the current users ip not
valid as a part of this change if there's another way.

ps zipkin-local-address is probably not a great name as it is for node
(maybe we'll figure out a browser based one?) perhaps name this
artifact based on the package name of the dependency?

adriancole:
@muymoo mind giving your opinion on this?

packages/zipkin/src/InetAddress.js
```diff
@@ -18,16 +18,4 @@ class InetAddress {
...
 
-// In non-node environments we fallback to 127.0.0.1
-InetAddress.getLocalAddress = function getLocalAddress() {
-  const isNode = typeof process === 'object' && typeof process.on === 'function';
-  if (!isNode) {
-    return new InetAddress('127.0.0.1');
-  }
-
-  // eslint-disable-next-line global-require
-  const networkAddress = require('network-address');
```
SimenB: if you do `require.apply(null, 'network-address')` you fool the react native bundler into not bundling this. as long as `isNode === false` above, that should fix the problem

SimenB: That should negate the need for an API change

DanielMSchmidt: Wow, thanks for the hints @SimenB, will try that!

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -3,13 +3,11 @@ const globalFetch =
   (typeof window !== 'undefined' && window.fetch) ||
   (typeof global !== 'undefined' && global.fetch);
 
-// eslint-disable-next-line global-require
-const fetch = globalFetch || require('node-fetch');
```
SimenB: try the `require`.call` workaround here as well? then I think you can avoid all api changes, right?

adriancole:
So anyone up to making the minimal changes to this branch such that https://github.com/openzipkin/zipkin-js-example/pull/5 works w/o browserify?

DanielMSchmidt:
Sorry, forgot to close this, I think we should continue the work in #126

PR #119, opened 2017-08-16
Kafka 010

### LINKED ISSUE #118: Upgrade to kafka-node@2
Might want to create a separate transport for people not using zookeeper

SimenB:
Fixes #118 

- First commit upgrades to `kafka-node@2`. The only breaking change is for consumers, and since we just produce here, that doesn't matter
  - https://github.com/SOHU-Co/kafka-node/blob/04a248ae1ffb4d6700823f77a80f84ed02948ba9/CHANGELOG.md#2017-07-08-version-200
- Second commit adds a `useZookeeper` option which defaults to `true` to avoid a breaking change
  - If set to `false` we create a new `KafkaClient` instead of `Client`, which does not use zookeeper
  - https://github.com/SOHU-Co/kafka-node/blob/04a248ae1ffb4d6700823f77a80f84ed02948ba9/README.md#kafkaclient

### DIFF
--- a/packages/zipkin-transport-kafka/src/KafkaLogger.js
+++ b/packages/zipkin-transport-kafka/src/KafkaLogger.js
@@ -14,9 +14,13 @@ module.exports = class KafkaLogger {
     const producerOpts = Object.assign({}, producerDefaults, options.producerOpts || {});
     this.producerPromise = new Promise((resolve, reject) => {
       this.topic = options.topic || 'zipkin';
-      this.client = new kafka.Client(
-        clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
-      );
+      if (clientOpts.connectionString) {
+        this.client = new kafka.Client(
+          clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
+        );
+      } else {
+        this.client = new kafka.KafkaClient(clientOpts);
+      }
       const producer = new kafka.HighLevelProducer(this.client, producerOpts);
       producer.on('ready', () => resolve(producer));
       producer.on('error', reject);
--- a/packages/zipkin-transport-kafka/test/integrationTest.js
+++ b/packages/zipkin-transport-kafka/test/integrationTest.js
@@ -112,4 +112,104 @@ describe('Kafka transport - integration test', () => {
       done(err);
     });
   });
+
+  it('should send trace data to Kafka without zookeeper', function(done) {
+    this.slow(10 * 1000);
+    this.timeout(60 * 1000);
+
+    makeKafkaServer().then(kafkaServer => {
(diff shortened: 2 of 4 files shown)


packages/zipkin-transport-kafka/test/integrationTest.js
```diff
@@ -112,4 +112,105 @@ describe('Kafka transport - integration test', () => {
       done(err);
     });
   });
+
+  it('should send trace data to Kafka without zookeeper', function(done) {
```
SimenB: This is copypasta except for `useZookeeper: false`. Might be prudent to make it DRYer?

packages/zipkin-transport-kafka/src/KafkaLogger.js
```diff
@@ -14,9 +15,15 @@ module.exports = class KafkaLogger {
...
-      this.client = new kafka.Client(
-        clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
-      );
+      if (useZookeeper) {
+        this.client = new kafka.Client(
+          clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
+        );
+      } else {
+        this.client = new kafka.KafkaClient(
+          Object.assign({kafkaHost: clientOpts.connectionString}, clientOpts)
```
adriancole: how about making clientOpts.kafkaHost instead of useZookeeper and hijacking connectionString as the broker host?

SimenB: Makes sense!

SimenB:
Changed to `kafkaHost`

packages/zipkin-transport-kafka/src/KafkaLogger.js
```diff
@@ -14,9 +14,13 @@ module.exports = class KafkaLogger {
...
       this.topic = options.topic || 'zipkin';
-      this.client = new kafka.Client(
-        clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
-      );
+      if (clientOpts.connectionString) {
+        this.client = new kafka.Client(
+          clientOpts.connectionString, clientOpts.clientId, clientOpts.zkOpts
+        );
+      } else {
+        this.client = new kafka.KafkaClient(clientOpts);
```
SimenB: way cleaner 😄

adriancole:
kicked the build in case it was a transient error

SimenB:
Green now :smile:

adriancole:
great job. thank you very much!

adriancole:
@SimenB would you mind looking at #117? We need more eyes on this, so that things can progress.

adriancole:
out in 0.8.0 thanks!

Issue #73, opened 2017-01-31
Feature request: GRPC client/server interceptors


devinsba:
It would be nice to have interceptors similar to the grpc ones from openzipkin/brave

devinsba:
Turns out there is a missing feature in the nodejs grpc library: https://github.com/grpc/grpc/issues/8767

adriancole:
since zipkin is very much like google cloud trace, we can probably borrow approach from them https://github.com/GoogleCloudPlatform/cloud-trace-nodejs

devinsba:
Their approach overall is quite nice, if I end up with time I'll take a crack at implementing it.

If someone else gets to it first: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/hooks/userspace/hook-grpc.js

ethanrubio:
👍  Would love to see this as well!

lnshi:
+1

adriancole:
https://github.com/grpc/proposal/pull/14 indicates a future interceptor model (which doesn't necessarily help now)

wdittmer:
I am using the code that @lnshi created already. I created a wrapper function that handles getting the metadata and finishing when the function is done. It would be nice if this could be included/done automatically when using this library. But it is a bit over my head at the moment.

```
function trace(func) {
    return async (call, callback) => {
        global.ZIPKIN_GRPC_INTCP.uponServerRecvGrpcCall({
            serviceName: ZIPKIN_SERVICE_NAME,
            grpcMetadataFromIncomingCtx: call.metadata
        });

        try {
            await func(call, callback);
        } finally {
            global.ZIPKIN_GRPC_INTCP.uponServerFinishRespond();
        }
    }
}
```

Then when assigning the call to the grpc server you just use `grpcMethod:  trace(your_original_grpc_method)`

your_original_grpc_method must be async, e.g.

```
async function your_original_grpc_method(call, callback) {
    try {
         // do stuff
         return callback(null, {message: 'message'});
    } catch (err) {
        return callback(err);
    }
}
```

jcchavezs:
I think this issue is partially fixes by https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-instrumentation-grpc-client. Ping @ghermeto

ghermeto:
it was... and the bad news is that gRPC in javascript doesn't have server interceptors (yet).

PR #121, opened 2017-08-22
Implement GRPC interceptor for Zipkin

lnshi:
- Implement GRPC interceptor for Zipkin.

- Addressed [Issue #73](https://github.com/openzipkin/zipkin-js/issues/73).

### DIFF
--- a/packages/zipkin-instrumentation-grpc/src/ZipkinGrpcInterceptor.js
+++ b/packages/zipkin-instrumentation-grpc/src/ZipkinGrpcInterceptor.js
@@ -1,7 +1,9 @@
 const grpc = require('grpc');
 
 const {
-  Annotation
+  TraceId,
+  Annotation,
+  option: {Some}
 } = require('zipkin');
 
 class ZipkinGrpcInterceptor {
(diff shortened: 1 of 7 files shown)


adriancole:
Thanks for raising this. Looks like most of the pieces are here. I'm wondering if you could package this into a shimmer module, perhaps with patchClient and patchServer functions like https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/plugins/plugin-grpc.js

This would be easier for folks to use as it would have less instructions

packages/zipkin-instrumentation-grpc/src/ZipkinGrpcInterceptor.js
```diff
@@ -0,0 +1,105 @@
...
+    let ctxSampled = grpcMetadataFromIncomingCtx.get('x-b3-sampled')[0];
+    if (!(ctxSampled in ['0', '1'])) {
+      ctxSampled = '0';
+    }
+
+    const ctxTraceInfo = new TraceId({
+      traceId: new Some(ctxTraceId),
+      parentId: new Some(ctxParentId),
+      spanId: ctxSpanId,
+      sampled: new Some(ctxSampled)
```
adriancole: per httpServer.js we shouldn't assume Some (ex if the header is not there). you can check for that one's impl as it is more correct

wdittmer: @lnshi 
Created a pull request that 'should' fix that: https://github.com/lnshi/zipkin-js/pull/1/commits/0931720f0aa1a170ad66ce35b52c949ca9c4544c

tomasAlabes:
Hi! @adriancole @lnshi is there a plan to merge this PR?

adriancole:
there's no plan to merge this pull request as-is as it has a bug and has gone dormant. If someone wants to raise a derived pull request, please do.

If someone wants to give it a retry, please simplify the README as currently it is hard to understand and includes instructions for golang which make things blurry. If you derive code from here, please credit @lnshi and try to simplify the setup steps as possible

PR #123, opened 2017-08-31
Swipe at clarifying usually there's one tracer

adriancole:


packages/zipkin/README.md
```diff
@@ -13,6 +13,7 @@ const CLSContext = require('zipkin-context-cls');
 const ctxImpl = new CLSContext(); // if you want to use CLS
 const xtxImpl = new zipkin.ExplicitContext(); // Alternative; if you want to pass around the context manually
 
+// Tracer will be a one to many relationship with instrumentation that use it (like express)
```
SimenB: How does this relate to `if you want to pass around the context manually` above?

SimenB: And I should have a single tracer for multiple instrumentations, right?

```js
const fetch = require('node-fetch');
const express = require('express');

const {Tracer, ExplicitContext, ConsoleRecorder} = require('zipkin');
const zipkinMiddleware = require('ipkin-instrumentation-express').expressMiddleware;

const ctxImpl = new ExplicitContext();
const recorder = new ConsoleRecorder();

const tracer = new Tracer({ctxImpl, recorder});

const zipkinFetch = wrapFetch(fetch, {tracer});
const app = express();

app.use(zipkinMiddleware({
  tracer,
  serviceName: 'server-1'
}));
```

adriancole: yes, multiple instrumentation would use the same tracer.

@eirslett Is it fair to say that if using explicit context, then you would need to use something like tracer.id()/setId() to address transitions across threads? (in any case you need to hold a reference to the tracer).

SimenB: Also, how does CLS relate to the newly released async_hooks API in node? (might deserve its own issue)

eirslett: > Is it fair to say that if using explicit context, then you would need to use something like tracer.id()/setId() to address transitions across threads? (in any case you need to hold a reference to the tracer).

I guess so...

> Also, how does CLS relate to the newly released async_hooks API in node? (might deserve its own issue).

We should either create a separate context implementation for async hooks, or update cls to use the async hooks mechanism (PR upstream I guess)

SimenB: Could there be an example in the docs of cases where tracer.id()/setId() is used to exemplify?

---

>We should either create a separate context implementation for async hooks, or update cls to use the async hooks mechanism (PR upstream I guess)

A separate package makes sense IMO, as async_hooks is still experimental (and is node8 only). I can open up a separate issue

SimenB: #124

PR #126, opened 2017-09-04
Defers access to libraries that require node.js

adriancole:
As started in #117, this defers or obviates loading libraries that only
exist in node.js, using a "require trick" which fools react native's
packager (as suggested by @SimenB).

This adds a little bit of debt inlining some thrift constants, but this
shouldn't be a maintenance burden as there are few thrift related
updates.


adriancole:
@DanielMSchmidt  @SimenB I was able to progress https://github.com/openzipkin/zipkin-js-example/pull/5 with these changes. Can one or both of you verify (nothing new is broken and that it works)

SimenB:
I can test it out tomorrow!

packages/zipkin-transport-http/package.json
```diff
@@ -11,14 +11,12 @@
...
-  "dependencies": {
-    "node-fetch": "^1.5.3"
-  },
   "devDependencies": {
-    "babel-cli": "^6.23.0",
+    "babel-cli": "^6.26.0",
     "body-parser": "^1.15.2",
     "express": "^4.13.4",
     "mocha": "^3.2.0",
+    "node-fetch": "^1.5.3",
```
SimenB: this should still be a dep?

adriancole:
I had to play games to stop react native from finding things, though
node_fetch might have been moved prematurely

packages/zipkin/src/serializeSpan.js
```diff
@@ -1,14 +1,16 @@
-const {TBufferedTransport, TBinaryProtocol} = require('thrift');
 const {fromByteArray: base64encode} = require('base64-js');
 
-let serialized;
-const transport = new TBufferedTransport(null, res => {
-  serialized = res;
-});
+module.exports = function serializeSpan(span, format = 'base64') {
+  // eslint-disable-next-line global-require
+  const {TBufferedTransport, TBinaryProtocol} = require.call(null, 'thrift');
```
DanielMSchmidt: serialize span is always called when you want to transport a span, right? This means as a non-node user you will require thrift, which will fail or am I mistaken?

adriancole:
>
> serialize span is always called when you want to transport a span, right?
> This means as a non-node user you will require thrift, which will fail or
> am I mistaken?
>
nope it is only used by kafka and scribe, I used this with the react thing,
but still want second or third eyes

DanielMSchmidt:
@adriancole Okay, I think the PR looks nice then, I will check out if it works 👍

packages/zipkin-transport-http/package.json
eirslett: maybe https://github.com/matthew-andrews/isomorphic-fetch is better, if it works with react-native?

adriancole:
>
> maybe https://github.com/matthew-andrews/isomorphic-fetch is better, if it works with react-native?
>
> started to try this, but it times out. seems some explicit "done" handling
needed

packages/zipkin-transport-http/package.json
SimenB: fetch-ponyfill doesn't modify the global environment, so that's probably a better alternative

adriancole:
there is something about the fetch polyfill stuf that makes tests die

```
  10) HTTP transport - integration test should send trace data via HTTP:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Here's the diff to switch out the fetch thing.. note that react did actually send data to zipkin before I added this diff!

```diff
diff --git a/packages/zipkin-transport-http/package.json b/packages/zipkin-transport-http/package.json
index 0b56212..0925739 100644
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -11,12 +11,14 @@
   "author": "Openzipkin <openzipkin.alt@gmail.com>",
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
+  "dependencies": {
+    "fetch-ponyfill": "^4.1.0"
+  },
   "devDependencies": {
     "babel-cli": "^6.26.0",
     "body-parser": "^1.15.2",
     "express": "^4.13.4",
     "mocha": "^3.2.0",
-    "node-fetch": "^1.5.3",
     "zipkin": "^0.7.3"
   }
 }
diff --git a/packages/zipkin-transport-http/src/HttpLogger.js b/packages/zipkin-transport-http/src/HttpLogger.js
index e734134..d7f4971 100644
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -1,10 +1,5 @@
 /* eslint-disable no-console */
-const globalFetch =
-  (typeof window !== 'undefined' && window.fetch) ||
-  (typeof global !== 'undefined' && global.fetch);
-
-// eslint-disable-next-line global-require
-const fetch = globalFetch || require.call(null, 'node-fetch');
+const fetch = require('fetch-ponyfill');
 
 class HttpLogger {
   constructor({endpoint, httpInterval = 1000}) {
```

SimenB:
I don't think the timeout is fetch's fault. If I put a breakpoint in the `logSpan` method, it's only called after 20 seconds. Something else changed in this PR which makes it fail

adriancole:
>
> I don't think the timeout is fetch's fault. If I put a breakpoint in the
> logSpan method, it's only called after 20 seconds. Something else changed
> in this PR which makes it fail
>
OK so maybe we can do a different PR on the fetch thing, huh. I'll do one
against master

DanielMSchmidt:
I build a small demo for React Native, but I don't have a public instance of zipkin running so I can't test it really. [Here is the demo](https://snack.expo.io/ry41Ny0K-)

DanielMSchmidt:
Hey there,

I double checked and react-native (and I also think most webpack based systems) are inlining the dependencies at compile time, which means only deferring them will not work. We need to move these requires to the call-side. I will try to pick up your work and move it as away as good as I can.

DanielMSchmidt:
I tried to move some of your efforts in a more finished way to #137, maybe you guys (@SimenB  @adriancole) can have a look

adriancole:
closed via #137

Issue #28, opened 2016-08-01
Feature request: typescript definitions


lukiano:
(if they don't already exist)

eirslett:
There are none yet, but it has been discussed. Probably a good idea! I have no experience with TypeScript.

adriancole:
These are in starting version 0.9. Though we do need tests somehow... imho (maybe we can write our normal tests in typescript?)

Issue #13, opened 2016-06-09
Input validation


eirslett:
We should make zipkin-js' public API fool-proof, by validating every input argument to every public function.

sveisvei:
Using typescript and generate typings could be one solution.
More low-hanging would be to use something like json-schema or joi if there are rich options objects.

PR #131, opened 2017-09-21
First go at typescript type definitions

pbadenski:
I started using this in our project. This is by no means complete and may contain some errors. But! It still helps immensely when adding custom integrations (if you use TypeScript that is).

Related to:
 - https://github.com/openzipkin/zipkin-js/issues/13
 - https://github.com/openzipkin/zipkin-js/issues/28

### DIFF
--- a/packages/zipkin/README.md
+++ b/packages/zipkin/README.md
@@ -3,6 +3,8 @@
 This is the core npm package for Zipkin. It contains the public API which is used by the various
 plugins (instrumentations and transports).
 
+We include TypeScript [definition file](index.d.ts) which you can also use a documentation.
+
 ## Usage
 
 ```javascript
--- a/packages/zipkin/README.md
+++ b/packages/zipkin/README.md
@@ -3,7 +3,11 @@
 This is the core npm package for Zipkin. It contains the public API which is used by the various
 plugins (instrumentations and transports).
 
-We include TypeScript [definition file](index.d.ts) which you can also use a documentation.
+We include TypeScript [definition file](index.d.ts) which you can also use as documentation.
+
+## Developing
+
+Please always make sure that [TypeScript type definitions](index.d.ts) match source code modifications.
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


adriancole:
looks nice. we should have something in the README and some tests to show this works (ex besides eyeballing which if the only option, that's ok)

pbadenski:
Makes sense. I'll have a shot at it.

pbadenski:
I added info to the README.

As for the example usage - for people who use TypeScript there's no setup, it's "you just use it" kind of thing. Only thing I can think of is - I could move **zipkin-instrumentation-redis** to TypeScript for the purpose of an example. Depends on how well this aligns with plans for **zipkin-js**.

P.S. I also fixed binary annotation type definition, based on my mess-up in #133

packages/zipkin/README.md
```diff
@@ -3,6 +3,8 @@
 This is the core npm package for Zipkin. It contains the public API which is used by the various
 plugins (instrumentations and transports).
 
+We include TypeScript [definition file](index.d.ts) which you can also use a documentation.
```
adriancole: 'use a documentation' or did you mean use as documentation?

packages/zipkin/index.d.ts
```diff
@@ -0,0 +1,159 @@
+declare namespace zipkin {
```
adriancole: Maybe make a comment when to change this? Ex whenever we change any file?

packages/zipkin/README.md
pbadenski: I did mean "use as documentation"

packages/zipkin/index.d.ts
pbadenski: I will add a comment.

pbadenski:
I understand your discomfort in this regard. Another idea - if tests were actually written in TypeScript, this would also help. But it does increase complexity. Unfortunately I'm not aware of other way to verify implementation matches TypeScript interfaces.

adriancole:
going out at 0.9.0 thanks for all the help, @pbadenski!

PR #116, opened 2017-07-27
Add support for tracing redis's 'batch' & 'multi'

pbadenski:

PR #133, opened 2017-09-28
Fix: Binary annotation cannot be array (must be string)

pbadenski:
In #116 I made an unfortunate error and passed an array to recordBinary.. This will create an erroneous trace, which cannot be sent to zipkin.


adriancole:
We have a failing test on redis due to possibly a different change. Are you able to have a look and help fix that broken window?

pbadenski:
Will do.

pbadenski:
It was my fault entirely! Fixed :D

adriancole:
thanks!

Issue #124, opened 2017-08-31
Feature request: async_hooks based context


SimenB:
Node 8 has a new async hooks module, which should be able to replace `zipkin-context-cls` entirely, as well as work better (i.e. at all) for promises.

It's still experimental, but not behind a flag. No polyfills exists for older nodes (lots of C++ in the implementation, so probably not incoming either).

https://nodejs.org/api/async_hooks.html

Obligatory hype blog post: https://medium.com/the-node-js-collection/async-hooks-in-node-js-illustrated-b7ce1344111f & 

/cc @sveisvei (https://github.com/openzipkin/zipkin-js/issues/65#issuecomment-268847034)

SimenB:
Might be worth taking a look at a bridge for cls: https://www.npmjs.com/package/cls-hooked

SimenB:
https://www.npmjs.com/package/asyncctx

anujshankar:
https://www.npmjs.com/package/zipkin-context-cls-hooked

Ghost---Shadow:
@anujshankar :)
#201 solves this

mziemer21:
Is there an approach that doesn't use async hooks? There currently are [performance issues](https://github.com/nodejs/benchmarking/issues/181) with async hooks.

jcchavezs:
@mziemer21 not at the moment. Actually this is a good input for this https://github.com/openzipkin/zipkin-js/pull/201#issuecomment-378436554

jcchavezs:
More relevant information: https://github.com/DataDog/dd-trace-js/issues/695

PR #135, opened 2017-10-04
context backed by cls-hooked (for async/await support)

pbadenski:
**zipkin-context-cls** does not support the new (> node v8) [async/await notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). This pull request addresses that.

Q: Why I haven't replaced **zipkin-context-cls** instead?
A: This library is backed by **cls-hooked** library which as I understand behaves differently for node < v8. I decided to be conservative and give the user (developer) a choice.

Fixes: #124

### DIFF
--- a/packages/zipkin-context-cls-hooked/README.md
+++ b/packages/zipkin-context-cls-hooked/README.md
@@ -1,0 +1,26 @@
+# zipkin-context-cls-hooked
+
+This module implements a context API on top of [Continuation-Local Storage ( Hooked )](https://github.com/jeff-lewis/cls-hooked).
+
+The primary objective of CLS is to implement a *transparent* context API, that is, you don't need to pass around a `ctx`
+variable everywhere in your application code.
+
+CLS-hooked is a fork of CLS to support **async/await** notation in Node v8 and higher.
+
+If using Node v7 or lower please make sure to understand the limitations of [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked)
+and consider using [zipkin-context-cls](https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-context-cls) instead.
+
+## Usage:
+
+```javascript
+const CLSHookedContext = require('zipkin-context-cls-hooked');
+const tracer = new Tracer({
+  ctxImpl: new CLSHookedContext('zipkin'),
+  recorder // typically Kafka or Scribe
+});
+```
+
+## A note on CLS context vs. explicit context
+
+There are known issues and limitations with CLS, so some people might prefer to use `ExplicitContext` instead;
+the drawback then is that you have to pass around a context object manually.
--- a/packages/zipkin-context-cls-hooked/package.json
+++ b/packages/zipkin-context-cls-hooked/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "zipkin-context-cls-hooked",
+  "version": "0.1.0",
+  "description": "A Context API implementation that uses cls-hooked under the hood",
+  "main": "lib/CLSHookedContext.js",
+  "scripts": {
+    "build": "babel src -d lib",
+    "test": "mocha --require ../../test/helper.js",
+    "prepublish": "npm run build"
+  },
+  "author": "OpenZipkin <openzipkin.alt@gmail.com>",
+  "license": "Apache-2.0",
+  "repository": "https://github.com/openzipkin/zipkin-js",
+  "dependencies": {
+    "cls-hooked": "^4.2.2"
+  },
+  "devDependencies": {
+    "babel-cli": "^6.23.0",
+    "mocha": "^3.2.0"
+  }
+}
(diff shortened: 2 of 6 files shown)


eirslett:
We could just do a 2.0 release where we remove babel, and only target node 8+?
If people are not ready to upgrade, they can use 1.x?

pbadenski:
IMHO it's a tad bit too soon. A sample report I found (from last year): https://semaphoreci.com/blog/2016/10/12/nodejs-versions-used-in-commercial-projects-2016-edition.html. Let's account for it being year old and ignore Node v4. Node v6 seems still to be very popular and node v8 LTS started only 4 days ago.

adriancole:
@SimenB verifying hooked works for you, right?

Seems like the dilemma here is async doesn't work without hooked and we need to test it somehow.

Is there a way to install v8 and make sure presence of hooked doesn't make unrelated tests fail?

eirslett:
We can conditionally skip tests in runtime, by checking the `process.versions.node` variable. But that won't work if the test includes `async`/`await` keywords - older versions of node.js won't be able to parse the code so it throws a SyntaxError before the conditional check can run.

Install semver dependency:

```
yarn install semver --dev
```

Check version:

```
// add a before/beforeEach hook to individual tests or test suites
before(function() {
  const minVersion = '8.0.0';
  if (semver.gt(process.versions.node, minVersion)) {
    console.log('Current version of node.js is ' + process.versions.node + '. Skipping tests that require node.js > ' + minVersion + '.');
    this.skip();
  }
});
```

Another variant is to have a separate `mocha.opts` file, which matches all `*-test-node8.js` and `*-test.js` files, so that the node8 tests files won't be matched by the default pattern. We would then have a `test-runner.js` that will check the Node.js version, choose the correct `mocha.opts` file, and run Mocha programmatically.

eirslett:
Yet another solution is to make `zipkin-context-cls-hooked` a separate contrib package on openzipkin-contrib, publish it separately, and bring it into the monorepo once the latest releases of the monorepo only support node.js 8+.

adriancole:
@eirslett not a bad idea wrt separate module in contrib. we should encourage this more I think

eirslett:
I still think the best solution is to release zipkin-js 1.0 which only supports node 8+. For all users of earlier versions of node.js, we will still have the 0.9 branch, with bugfixes and patches (we could LTS it).

I think it's reasonable to ask users to use zipkin-js 0.9 if they're on an older version of node.js. The 0.9 branch is stable and feature complete, it has all the packages you would commonly need. (Except for new instrumentation plugins)

Given the end user risk of upgrading node.js to 8: Users who need new features that are not in 0.9, are either probably introducing zipkin to a greenfield project or making major changes to an existing brownfield project to introduce zipkin - both are high-risk operations, where an upgrade of node.js to 8 wouldn't be the most risky part of it.

If users are unable to upgrade to node 8, and they really need a new instrumentation package, they could either publish a separate one on their own npm repo or openzipkin-contrib, or upgrade to node.js 8.

Even if an instrumentation package were to exist only in zipkin-js 1.x, chances are it would work on node 6 anyways, if people use zipkin-context-cls instead of the cls-hooked. (Given that we don't make breaking API changes in core from 0.9 to 1.0)

SimenB:
It seems weird this single module should be excluded from the monorepo.

>I still think the best solution is to release zipkin-js 1.0 which only supports node 8+.
Node 8 will be LTS sometime this month

adriancole:
there are a fair number of zipkin-js npm packages published outside the monorepo, just usually they aren't proposed here first. regardless, indeed we should consider points mentioned and make a call.

adriancole:
PS readying for a release that will need to go out prior to this. One thing I'd like to clarify that this doesn't in any way affect the support of in-browser or react native stuff, it just limits the node environments supported, right?

pbadenski:
I'm not sure actually, I'll check that.

adriancole:
ping @pbadenski in case you aren't away for holidays.. thoughts on the last comment?

pbadenski:
@adriancole thanks for reminder, not on holidays, just been insanely busy recently. I am aware this and related pull requests have been stale for quite long now... Based on all conversations I think this issue is still dependent on #146. I'll try to move it further along this week (but also feel free anyone to take over from me).

Based on reading `cls-hooked` code it won't work with browser out of the box - due to dependencies on various node libraries (eg. **util** & **assert**). If we replace `zipkin-cls-context` to use `cls-hooked` we can either:
 - rename and keep current `zipkin-cls-context` as fallback for browser/native,
 - or write some fancy code that selects different dependency for node vs browser.

adriancole:
Looking at this, I feel like we could make zipkin-cls-context-hooked a separate "subdirectory", maybe this one, but unhooked from the main build. This removes the problem about compat with coupling of native and browser code. As long as there's no api affecting concerns, CLS code is rarely ever bumped anyway. Our README could have a separate step for the rare time we release it, and we wouldn't need to decomission a repo when the world converges, rather just hook it to the main build.

For CI, we could setup circleci and have it run this module alone. WDYT?

jcchavezs:
@pbadenski any chance you can restart this great work?

pbadenski:
Unfortunately I won't be able to pick this up. We're not using zipkin in the project any more and I don't really have time otherwise. Sorry.

PR #136, opened 2017-10-04
Helpers for instrumenting http clients & servers

pbadenski:
Removes duplication and provides an easier/quicker way to build zipkin instrumentation for http clients & servers.

Conceptually inspired by [brave instrumentation helpers](https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpServerHandler.java#L94) (thanks to @adriancole for the tip)


packages/zipkin/src/instrumentation/httpClient.js
```diff
@@ -0,0 +1,46 @@
...
+    tracer = requiredArg('tracer')
+  }) {
+    this.tracer = tracer;
+  }
+
+  recordRequest(serviceName, request, remoteServiceName, url, method) {
+    this.tracer.setId(this.tracer.createChildId());
+    const traceId = this.tracer.id;
+
+    this.tracer.recordServiceName(serviceName);
```
adriancole: serviceName should be constant for the instrumentation.. move up to the ctor?

packages/zipkin/src/instrumentation/httpClient.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class HttpClientInstrumentation {
+  constructor({
+    tracer = requiredArg('tracer')
+  }) {
+    this.tracer = tracer;
+  }
+
+  recordRequest(serviceName, request, remoteServiceName, url, method) {
```
adriancole: the remoteServiceName can also be in the ctor (as this won't change per request even if the associated IP can)

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -0,0 +1,89 @@
...
+    this.tracer.recordAnnotation(new Annotation.ServerRecv());
+    this.tracer.recordAnnotation(new Annotation.LocalAddr({port}));
+
+    if (id.flags !== 0 && id.flags != null) {
+      this.tracer.recordBinary(Header.Flags, id.flags.toString());
+    }
+    return id;
+  }
+
+  recordResponse(id, statusCode) {
```
adriancole: seems recordResponse should also have overload for error, or a second optional arg for error

pbadenski: Will do. It does make sense (even though no http server instrumentation libraries currently make any attempt to record errors). I'd be leaning towards adding extra parameter here since it's a server and most likely will return 5xx in addition to that error.

packages/zipkin/src/instrumentation/httpClient.js
pbadenski: 👍

packages/zipkin/src/instrumentation/httpClient.js
pbadenski: 👍

adriancole:
still looks like progress to me. Some internal details will flex a little, but great.

PR #137, opened 2017-10-10
Support for React Native and Web

DanielMSchmidt:
With this PR merged I could get the React Native example to work.
I don't know if the tests still work, but this first try should show roughly how it can work.


DanielMSchmidt:
@adriancole Would you have any time and patience to look into this? I don't understand the errors, as I don't fully understand what happens, but I can say that I got spans into zipkin from my react-native app running in a simulator

adriancole:
I'm probably more inclined to see a zipkin-encoder-thrift module (somehow move the toThrift functionality to there, too). In theory seems a nice way to move the whole functionality vs a piece of it. Do you have cycles to try that out?

DanielMSchmidt:
Could we go there incrementally? I can give it a try but I would love to see a working solution first and iterate on it

adriancole:
@DanielMSchmidt the problem is that if we merge this, we pollute npm with a temporary package.. I'm ok with incremental, just it locks us from doing a release.

DanielMSchmidt:
You are right, I didn't think about that. I will make the move you
suggested the next days. We could also do a relative import for the
serialization, as we produce a transpiled version anyhow, what do you think?

Adrian Cole <notifications@github.com> schrieb am Fr., 13. Okt. 2017, 07:42:

adriancole:
>
> We could also do a relative import for the
> serialization, as we produce a transpiled version anyhow, what do you
> think?
>
>
I'd need someone else to weigh in on this option as I have no clue :P

DanielMSchmidt:
Differently posed: is there any value in adding a new package?

Adrian Cole <notifications@github.com> schrieb am Fr., 13. Okt. 2017, 10:22:

DanielMSchmidt:
@adriancole I moved the logic into a share `internals/` folder, it shouldn't matter as the requires are evaluated at compile time, afaik

eirslett:
I don't think it should be done incrementally, just do it in one PR.

DanielMSchmidt:
@eirslett i already pushed the changes needed, but I don't really get why the build fails

packages/zipkin-transport-kafka/src/KafkaLogger.js
```diff
@@ -1,5 +1,5 @@
 const kafka = require('kafka-node');
-const {serializeSpan} = require('zipkin');
+const serializeSpan = require('../../../internals/serialize-span');
```
adriancole: how does the "internals" approach for those using npm? Is there a specific project setup they would need?

DanielMSchmidt: I was under the assumtion it would be inlined, but it isn't. I will do some research how to do it or if it's impossible I will just create a symlink 👍  [preval](https://github.com/kentcdodds/babel-plugin-preval) might be an option for inlining 🤔

eirslett: Symlinks don't work on Windows

DanielMSchmidt:
I had problems with the other incremental steps and I think you were right @adriancole 

I will try my best to extract the right abstraction level.

adriancole:
Thanks for the continued effort!

DanielMSchmidt:
@adriancole Thank you! Your effort in handling PRs is highly appreciated 👍 

I wonder what else besides the serialize span functionality should be inside the package you would like to have

adriancole:
I think we'd only want thrift-related types and dependencies and the serialize function. Note: that function was made when there was only one way to serialize, so should do a reality check it makes sense to export across packages. second reality check is trying the sample apps to make sure (ex using something like scribe or kafka to ensure thrift encoding is used)

idanhaviv:
Hi, I tried to consume this PR with "git+https://github.com/openzipkin/zipkin-js.git#c0780cb7359a19105559b117d2014f937842beb9" as zipkin dependency in package.json with no luck. I got an error "No version provided in package.json". i'm guessing it has to do with Lerna, also tried adding another '/zipkin' to the path, github user and password, and still failed. any idea how can i use it?

DanielMSchmidt:
@idanhaviv You are right, won't work because of lerna. You might clone it locally and use `lerna bootstrap` and `npm link` for the packages you need afterwards as a temporary fix, but let's try to get this PR merged 👍

DanielMSchmidt:
@adriancole Can we somehow get this merged? I fixed the eslint issues, but I don't understand the failing tests. Also I don't see a point in not moving this forward. I had a look into our thrift dependencies and I think they are kinda hard to abstract away

adriancole:
I agree this is useful. Will look at the failing tests. Thanks for the help.

adriancole:
FYI I'm thinking about an executive decision to just copy/paste the thrift dep to the two call sites. The reason is that there are no more transports that will accept thrift in the future, and this way we don't need to expose any temporary modules (especially ones not relevant in v2 format)

scream if you hate the idea :P

adriancole:
hmm actually we still have the thrift types in the core module... I will try to clean this up (timeboxing a bit)

DanielMSchmidt:
@adriancole thanks for taking the time. It seemed to me that thrift is a pretty central part, although we don't need most of it. Are you aware that there is a web version of thrift, not only a node one? Could that be an option?

adriancole:
I think the cleanest way is to add the v2 span type then expose that.. then
zipkin-encoder-thrift can use it to create the v1 thrift encoding. v2 span
is almost the same as MutableSpan

adriancole:
I will be offline for about 2hrs, but if you want to give a crack at
it, here is the Span type
http://zipkin.io/zipkin-api/#/default/post_spans

should look something like this in typescript

  type Endpoint = {
    'serviceName': string
    'ipv4': string
    'ipv6': string
    'port': number
  };

  type Annotation = {
    'timestamp': number
    'value': string
  };

  type Span = {
    'traceId': string
    'parentId': string
    'id': string
    'kind': "CLIENT" | "SERVER" | "PRODUCER" | "CONSUMER"
    'name': string
    'timestamp': number
    'duration': number
    'localEndpoint': Endpoint
    'remoteEndpoint': Endpoint
    'annotations': Array < Annotation >
    'tags': {}
    'debug': boolean
    'shared': boolean
  };

Notably, we'd replace MutableSpan with "Span" based on above.
binaryAnnotations -> tags (though stringified), then instead of "cs"
it would be kind=CLIENT, timestamp=a.timestamp  and "cr" would set
duration to a.timestamp - span.timestamp

then, all of the thrift code can move to zipkin-encoder-thrift (all of
it including the codegen stuff and the tests)

adriancole:
Ps whatever the replacement for mutablespan is, probably shouldnt be named
Span exactly as Span will become an api at some point. Maybe SpanData (ex
census uses this to differentiate a record from an api but open to ideas)
regardless the loggers aka reporter aka exporter to zipkin should accept
whatever we name it and encode afterwards

adriancole:
ok I squashed prior work into one commit and then redoing things to extract span encoding logic into another one. I'm pretty tired and could use a hand reviewing the latter commit. basically it renames MutableSpan to record.Span and makes it almost the same as span v2. Right now, there's only v1 encoding, but this work can merge and we can make it possible to do v2 later.

also kudos if someone wants to help de-lint this :)

gotta sleep, but hope this helps

packages/zipkin-encoder-thrift/src/index.js
```diff
@@ -0,0 +1,113 @@
...
+  } else {
+    res.trace_id_high = traceId.substr(0, 16);
+    res.trace_id = traceId.substr(traceId.length - 16);
+  }
+  res.name = span.name.getOrElse('unknown');
+  const thriftEndpoint = toThriftEndpoint(span.endpoint, span.service);
+
+  let addressKey = 'sa';
+  res.annotations = span.annotations.map(ann => {
+    // In the RPC span model, the client owns the timestamp and duration of the
```
adriancole: after this change... not too long from now, I'll clean up this which is repeated and has been replaced with "span.shared" present in span v2

packages/zipkin/src/record.js
```diff
@@ -0,0 +1,63 @@
...
+  this.endpoint = ep;
+};
+Span.prototype.toString = function toString() {
+  const annotations = this.annotations.map(a => a.toString()).join(', ');
+  return `Span(id=${this.traceId.toString()}, annotations=[${annotations}])`;
+};
+
+module.exports.Endpoint = Endpoint;
+module.exports.Annotation = Annotation;
+module.exports.Span = Span;
```
adriancole: fyi @narayaruna this will make the "Span" type visible for future work, such as fancier batching. This type is almost v2 now, but I don't want to do the full v2 conversion in this pull request, as this one is about making thrift completely optional

packages/zipkin-encoder-thrift/test/toThrift.test.js
```diff
@@ -0,0 +1,196 @@
...
+  TBinaryProtocol
+} = require('thrift');
+const {fromByteArray: base64encode} = require('base64-js');
+const {
+  TraceId,
+  record: {Span, Endpoint},
+  option: {Some, None}
+} = require('zipkin');
+
+describe('Thrift Formatting', () => {
```
adriancole: this was a bear to make, but I noticed some logic only existed in json codec not thrift. I backfilled the tests

DanielMSchmidt:
I didn't get eslint to work 100%, just one error was still there.

adriancole:
>
> I didn't get eslint to work 100%, just one error was still there.
>
all help appreciated!

adriancole:
@DanielMSchmidt just a little polishing left. If you can try out the work in progress, would be nice.

adriancole:
whoot green! (appveyor is dying on installing yarn for some reason)

DanielMSchmidt:
Nice, great job 👍 I will test this implementation with RN today!

adriancole:
ok timestamp logic is sorted (though there's an eslint problem I'm not sure how to address)
only one thing left after this (use span.kind), so if it works for you we can cut something this weekend maybe tomorrow.

appreciate a hand with the eslint thing as otherwise will take me a while to figure it out. I'll be backonline in a few hours

DanielMSchmidt:
@adriancole I will try to get eslint to work, would be cool to have it today or tomorrow, two of my projects depend on this being merged 🙈

eirslett:
> I will also test if it works with react native, so that we are sure no refactoring broke it

I think we need integration tests to run on Travis, to make sure we don't break React Native. You know, people won't test React Native manually for every commit/release.

adriancole:
Agreed. Anyone know a project that has react integration tests setup?
Searching on github often is fruitful.. help wanted!

Ps i will revisit in about 13hrs or so.. but i cant work all weekend on
this (maybe i can spend another 2 hrs or so)

if someone can help with integration tests and the lint thing between now
and then, a release this weekend becomes significantly more likely!

On 27 Oct 2017 7:18 pm, "Eirik Sletteberg" <notifications@github.com> wrote:

I will also test if it works with react native, so that we are sure no
refactoring broke it

I think we need integration tests to run on Travis, to make sure we don't
break React Native. You know, people won't test React Native manually for
every commit/release.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/zipkin-js/pull/137#issuecomment-339944571>,
or mute the thread

DanielMSchmidt:
@adriancole @eirslett I am working on something, if we get this merged I will do a PR. Got a start point here in this repo: https://github.com/react-tracing/react-tracing

DanielMSchmidt:
Just to make this clear, I will build a small example as part of my work in `react-tracing` (which I will do as soon as you merge this PR) and afterwards I will share this basic pipeline setup and the E2E tests with you (only react so far, as the react-native E2E testing tool I work with currently doesn't support running in a CI 100%) As I work in this `react-naitve` area I will try to move the work upstream 👍

DanielMSchmidt:
Would it be okay for you if I did a seperate PR for a `node` e2e test and a `web` e2e test, adding the `react-native` one later on? I will also try to fix the eslint issue, but honestly, I am not willing to put in any more effort into this PR; I have the feeling it becomes kind of a rabit hole

adriancole:
>
> Would it be okay for you if I did a seperate PR for a node e2e test and a
> web e2e test, adding the react-native one later on?
>
yes. I trust you will later, as you've been very responsive on other
things. thanks!

no worries about the eslint, I will rewrite the code so that it passes.

adriancole:
Fyi I have everything sorted locally wrt lint etc

adriancole:
OK ready to merge on my end, but I do need someone to verify it.

packages/zipkin/index.d.ts
```diff
@@ -150,8 +150,6 @@ declare namespace zipkin {
     record: (rec: any) => void;
   }
 
-  const serializeSpan: () => void;
```
adriancole: ps a later request can formalize the typescript of accessing encoders.. we shouldn't do it, yet as for example the result isn't standard. Specifically, json returns a string whereas thrift returns a Buffer. That's the long story why I didn't replace this with anything.. we'll revisit later and do it more formally

packages/zipkin/index.d.ts
```diff
@@ -150,8 +150,6 @@ declare namespace zipkin {
...
-
   class ExplicitContext {
```
adriancole: if someone wants to add typescript for model.Span or model.Endpoint, they could as well.. we can also wait until we add the v2 api (as most won't work with model types directly)

packages/zipkin/src/batch-recorder.js
```diff
@@ -1,11 +1,13 @@
...
-const {
-  MutableSpan,
-  Endpoint,
-  ZipkinAnnotation,
-  BinaryAnnotation
-} = require('./internalRepresentations');
+const {now, hrtime} = require('./time');
+const {Span, Endpoint} = require('./model');
+
+function PartialSpan(traceId) {
```
adriancole: ended up aligning name with partial span which is used elsewhere in the same file

adriancole:
I went ahead and finished the job, notably http logger now can use the v2 json format.

Will merge once someone besides me says so

DanielMSchmidt:
Awesome @adriancole Can I help you in any way to get this PR merged? I opened #141 but I think I need some help. I currently can't get my test to pass, I think I am misusing zipkin-js somehow. I am sure your trained eyes may see my error immediately 👍

adriancole:
I merged it as you'll notice.. I'll mention some details on the other thread about E2E

DanielMSchmidt:
Awesome @adriancole, thanks for taking the effort! Can you do a release? This would be super awesome 👍

adriancole:
just tried a release and I think it pushed everything except the thrift
encoder package. there's an access issue I'm trying to work out cc @abesto

adriancole:
ok should be all good now

DanielMSchmidt:
Awesome, great job 👍

adriancole:
thanks for your help and the continuing stuff. If you can update the react
example over https://github.com/openzipkin/zipkin-js-example/pull/7 would
be apprecated

adriancole:
those wanting to try this should use https://github.com/openzipkin/zipkin-js-example which now works cc @idanhaviv

PR #141, opened 2017-10-28
WIP: E2E tests 

DanielMSchmidt:
This PR aims to add E2E tests for the node and web platform. The goal is to write the very first simple E2E tests, e.g. logging a span and querying it in the openzipkin docker instance.

Current status:

- [ ] node
    - [X] travis setup
    - [ ] tests
- [ ] web
    - [x] travis setup
    - [ ] tests


adriancole:
one of the jobs passed. another had this error: " missing script: lerna-bootstrap"

adriancole:
fyi just verified 0.10.0 manually here https://github.com/openzipkin/zipkin-js-example/pull/7

DanielMSchmidt:
Sorry for the delay, had problems wrapping my head around this. I think it might make sense to not test against a real zipkin server as the sync is really problematic. I will give it a new try with running an express server that mocks responses and that we can introspect. The solution should look similar to the one taht is done in [voldern/zipkin-instrumentation-got](https://github.com/voldern/zipkin-instrumentation-got)

DanielMSchmidt:
I think this approach is not really feasible, it is kinda hard to get it all together with lerna and the different environments we need to support. What I could do is write an example react & react native application that ensures when updated to the latest version that nothing is severely broken, what do you think @adriancole?

adriancole:
sounds fine to me, Daniel. We could put CI on the example repo as well.

adriancole:
I think @jcchavezs has some experience with E2E tests in the PHP repo. Would make sense to use a similar approach I think. For now, closing this while certainly a good reference.

PR #142, opened 2017-10-29
Cleans up glitches, notably re-enabling 'unknown' service names

adriancole:
Integrating react native, I noticed we weren't recording spans against
the default service name set by instrumentation (always 'unknown').

This also fixes some other glitches noticed when looking carefully at
the data as well the lint problem ignored formerly.

### DIFF
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -6,8 +6,14 @@ function PartialSpan(traceId) {
   this.startTimestamp = now();
   this.startTick = hrtime();
   this.delegate = new Span(traceId);
-  this.localEndpoint = new Endpoint({serviceName: 'unknown'});
+  this.localEndpoint = new Endpoint({});
 }
+PartialSpan.prototype.finish = function finish() {
+  if (this.endTimestamp) {
+    return;
+  }
+  this.endTimestamp = now(this.startTimestamp, this.startTick);
+};
 
 class BatchRecorder {
   constructor({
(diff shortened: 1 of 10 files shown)


adriancole:
@DanielMSchmidt this is why you weren't seeing spans from react. all else was good when I changed the default serviceName to something else

packages/zipkin/src/batch-recorder.js
```diff
@@ -6,8 +6,14 @@ function PartialSpan(traceId) {
   this.startTimestamp = now();
   this.startTick = hrtime();
   this.delegate = new Span(traceId);
-  this.localEndpoint = new Endpoint({serviceName: 'unknown'});
+  this.localEndpoint = new Endpoint({});
```
adriancole: the batch recorder shouldn't be setting this.. it should always be explicitly set by either instrumentation or the tracer (otherwise there's a difference in behavior when reading the console logger vs the batch recorder)

packages/zipkin/src/batch-recorder.js
```diff
@@ -64,17 +71,6 @@ class BatchRecorder {
     return span.startTimestamp + this.timeout < now();
   }
 
-  _decorateEndpoint(endpoint, ann) {
-    /* eslint-disable no-param-reassign */
```
adriancole: adding setter methods was the key

packages/zipkin/test/batch-recorder.test.js
```diff
@@ -142,7 +142,7 @@ describe('Batch Recorder', () => {
     });
   });
 
-  it('should record minimum duration of 1 microsecond', () => {
```
adriancole: minimum duration is now checked in model.test.js as it has nothing to do with this type

adriancole:
@DanielMSchmidt this is why you weren't seeing spans from react. all else was good when I changed the default serviceName to something else

adriancole:
ps one of us should fix appveyor soon :)

adriancole:
0.10.1 on the way. will retest sample apps

PR #144, opened 2017-10-30
Update type definitions

pbadenski:
Mostly updating after #137 and #142 changes

Additionally add:
 - sampler namespace
 - Record interface
 - TraceId properties

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -14,6 +14,19 @@ declare namespace zipkin {
     letContext<V>(ctx: T, callback: () => V): V;
   }
 
+  namespace sampler {
+    class Sampler {
+      constructor(evaluator: (traceId: TraceId) => boolean);
+    }
+
+    class CountingSampler implements Sampler {
+      constructor(sampleRate?: number);
+    }
+
+    const neverSample: (traceId: TraceId) => boolean;
+    const alwaysSample: (traceId: TraceId) => boolean;
+  }
+
   class Tracer {
     constructor(args: { ctxImpl: Context<TraceId>, recorder: Recorder, sampler?: sampler.Sampler, traceId128Bit?: boolean });
     id: TraceId;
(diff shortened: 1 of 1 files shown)


packages/zipkin/index.d.ts
```diff
@@ -85,9 +112,48 @@ declare namespace zipkin {
     }
   }
 
+  namespace model {
+    interface Endpoint {
+      constructor(args: { serviceName?: string, ipv4?: InetAddress, port?: number });
+
+      setServiceName(serviceName: string): void;
+      setIpv4(ipv4: InetAddress): void;
```
adriancole: technically this is a string literal which has an ipv4 inside..

packages/zipkin/index.d.ts
```diff
@@ -85,9 +112,48 @@ declare namespace zipkin {
...
+      traceId:        option.IOption<string>;
+      spanId:         string;
+      name:           string;
+      kind:           string;
+      timestamp:      number;
+      duration:       number;
+      localEndpoint:  Endpoint;
+      remoteEndpoint: Endpoint;
+      annotations:    any[];
+      tags:           any;
```
adriancole: is there a way to say dict string -> string?

packages/zipkin/index.d.ts
```diff
@@ -85,9 +112,48 @@ declare namespace zipkin {
...
+      id:             string;
+      traceId:        option.IOption<string>;
+      spanId:         string;
+      name:           string;
+      kind:           string;
+      timestamp:      number;
+      duration:       number;
+      localEndpoint:  Endpoint;
+      remoteEndpoint: Endpoint;
+      annotations:    any[];
```
adriancole: we could consider exporting model.Annotation I guess..

packages/zipkin/index.d.ts
pbadenski: yes: ```{ [key: string]: string }```. Will do that.

packages/zipkin/index.d.ts
pbadenski: Makes sense. Will do.

packages/zipkin/index.d.ts
pbadenski: oh, ipv4 is a string (of a "X.X.X.X" format)?

adriancole:
Yep this model type matches exactly the openapi/swagger def which is ip
string literal. We resolve from InetAddress type prior.

packages/zipkin/index.d.ts
```diff
@@ -117,12 +117,17 @@ declare namespace zipkin {
...
       setServiceName(serviceName: string): void;
-      setIpv4(ipv4: InetAddress): void;
+      setIpv4(ipv4: string): void;
       setPort(port: number): void;
 
       isEmpty(): void;
     }
 
+    interface Annotation {
+      timestamp: string;
```
adriancole: number, right?

adriancole:
rebased and force pushed.. PTAL

pbadenski:
👍 Looks good to me. Thanks!

PR #148, opened 2017-11-03
Add support for setting custom HTTP headers in HTTP transport

voldern:
This is needed because our Zipkin server is protected by tokens that must be sent as HTTP headers.

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -17,7 +17,6 @@ class HttpLogger {
     this.jsonEncoder = jsonEncoder;
 
     this.headers = Object.assign({
-      Accept: 'application/json',
       'Content-Type': 'application/json'
     }, headers);
 
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -11,10 +11,15 @@ const {
 } = require('zipkin');
 
 class HttpLogger {
-  constructor({endpoint, httpInterval = 1000, jsonEncoder = JSON_V1}) {
+  constructor({endpoint, headers = {}, httpInterval = 1000, jsonEncoder = JSON_V1}) {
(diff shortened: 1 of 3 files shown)


voldern:
Thanks for the review @adriancole.

I've updated the tests. 

Didn't find an issue though. Do you want me to make a new one and associate it with this PR?

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -11,11 +11,16 @@ const {
...
 
 class HttpLogger {
-  constructor({endpoint, httpInterval = 1000, jsonEncoder = JSON_V1}) {
+  constructor({endpoint, headers = {}, httpInterval = 1000, jsonEncoder = JSON_V1}) {
     this.endpoint = endpoint;
     this.queue = [];
     this.jsonEncoder = jsonEncoder;
 
+    this.headers = Object.assign({
+      Accept: 'application/json',
```
adriancole: nit but we can save some overhead I think by nixing the Accept. it isn't usually required especially as we aren't parsing anything (and the server only returns json)

adriancole:
I'd update the README with a subsection at the bottom showing how to add headers for the purpose of an guarded endpoint. Otherwise LGTM.. we've had similar requests in different libraries. For example, you can do the same in the java variant by accessing the http connection builder https://github.com/openzipkin/zipkin-reporter-java/blob/8a3ae3f7be93101b3db0841708c28eeb3ab7eebb/okhttp3/src/main/java/zipkin2/reporter/okhttp3/OkHttpSender.java#L88

packages/zipkin-transport-http/src/HttpLogger.js
voldern: :white_check_mark:

voldern:
> I'd update the README with a subsection at the bottom showing how to add headers for the purpose of an guarded endpoint.

I ended up documenting all options to the HTTP transport in the example.

Issue #149, opened 2017-11-05
Postgres instrumentation


voldern:
It would be great to have an instrumentation for [node-postgres](https://www.npmjs.com/package/pg).

Has anyone implemented this privately?

If not I can take a stab at it. The only methods that need wrapping seem to be [pg.Pool.query](https://node-postgres.com/api/pool#pool-query) and [pg.Client.query](https://node-postgres.com/api/client#client-query).

adriancole:
You did a great job. We do have a trend of rule-of-three adding instrumentation.. I wonder if you know any others who would use this? cc @tobespc

adriancole:
we've met rule-of-three. merging!

voldern:
Thanks for the help @adriancole :bowing_man:

PR #150, opened 2017-11-07
Add postgres instrumentation

voldern:
Resolves https://github.com/openzipkin/zipkin-js/issues/149

### DIFF
--- a/packages/zipkin-instrumentation-postgres/test/integrationTest.js
+++ b/packages/zipkin-instrumentation-postgres/test/integrationTest.js
@@ -1,0 +1,166 @@
+const sinon = require('sinon');
+const {Tracer, ExplicitContext, InetAddress} = require('zipkin');
+const zipkinClient = require('../src/zipkinClient');
+
+const postgresConnectionOptions = {
+  host: 'localhost',
+  port: 5432,
+  database: 'postgres',
+  user: 'postgres',
+  password: process.env.POSTGRES_PWD
+};
+
+const Postgres = require('pg');
+
+function getPostgres(tracer) {
+  return zipkinClient(tracer, Postgres);
+}
+
+function expectAnnotationsDescribePostgresInteraction(annotations) {
+  const sn = annotations[1].annotation;
+  expect(sn.annotationType).to.equal('ServiceName');
+  expect(sn.serviceName).to.equal('unknown');
+
+  const sa = annotations[2].annotation;
+  expect(sa.annotationType).to.equal('ServerAddr');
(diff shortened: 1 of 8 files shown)


packages/zipkin-instrumentation-postgres/test/integrationTest.js
```diff
@@ -0,0 +1,166 @@
...
+    const postgres = new (getPostgres(tracer).Client)(postgresConnectionOptions);
+    postgres.connect();
+
+    postgres.query('INVALID QUERY', (firstError) => {
+      postgres.query('ERROR QUERY').catch((secondError) => {
+        const query = new Postgres.Query('FAILED QUERY()');
+        const result = postgres.query(query);
+        result.on('error', (thirdError) => {
+          const annotations = recorder.record.args.map(args => args[0]);
+          expectErrorAnnotation(annotations[5], firstError);
```
adriancole: maybe describe what this is doing here? ex are these errors in different spans? (ps 'error' is often a binary annotation aka tag)

packages/zipkin-instrumentation-postgres/test/integrationTest.js
```diff
@@ -0,0 +1,166 @@
...
+          runTest(annotations.slice(10, 15));
+
+          expect(
+            annotations[0].traceId.spanId
+          ).not.to.equal(annotations[5].traceId.spanId);
+          expect(
+            annotations[0].traceId.spanId
+          ).not.to.equal(annotations[10].traceId.spanId);
+
+          annotations.forEach(ann => {
```
adriancole: style, but I'd find this stuff easier to understand if in span format vs a list of annotations. For example, you can look at batch-recorder.test.js

voldern: Based it on https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-memcached/test/integrationTest.js.

But this approach seems much clearer and succinct. Updated tests now.

packages/zipkin-instrumentation-postgres/test/integrationTest.js
voldern: Good catch! Fixed it.

adriancole:
@brianc fyi in zipkin-land we usually merge things after there's a few folks interested. This is a client-side trace integration (using your library). What this means is that postgres time and metadata  will end up included in traces that originate from anywhere (even an upstream browser request). Just in case you are interested.

PR #156, opened 2017-11-10
Adds Tracer.local for local tracing

### LINKED ISSUE #104: local tracing
There are a number of zipkin libraries supporting local tracing (operations that spawn in-process and are not RPCs).

We've recently had a request by @wirehead for how to do this.

```javascript
// Is this how I'm supposed to instrument a long-running function?
function ExampleTraceFunction() {
  tracer.setId(tracer.createRootId());
  tracer.recordMessage('start');
  // do stuff here
  tracer.recordMessage('end');
}
```

## workaround
I suggested a workaround, which is to use Client annotations while we figure out how we want to address this here.

```javascript
function ExampleTraceFunction() {
  tracer.scoped(() => {
    tracer.setId(tracer.createRootId());
    tracer.recordAnnotation(new Annotation.ClientSend());

    // do work

    tracer.recordAnnotation(new Annotation.ClientRecv());
  });
}
```

There's a background issue in finagle which might steer implementation. https://github.com/twitter/finagle/issues/541

cc @fedj

adriancole:
Inspired by the same design in https://github.com/openzipkin/zipkin4net 

Fixes #104

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -34,6 +34,7 @@ declare namespace zipkin {
     id: TraceId;
 
     scoped<V>(callback: () => V): V;
+    local<V>(name: string, callback: () => V): V;
     createRootId(): TraceId;
     createChildId(): TraceId;
     letChildId<V>(callback: (traceId: TraceId) => V): V;
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -90,6 +90,7 @@ class Tracer {
     return this.scoped(() => {
       const traceId = this.createChildId();
       this.setId(traceId);
+      this.recordServiceName(this._localEndpoint.serviceName);
       this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
 
       let result;
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -80,14 +80,20 @@ class Tracer {
     return childId;
   }
 
+  // creates a span, timing the given callable, adding any error as a tag
   local(operationName, callable) {
     return this.scoped(() => {
       const traceId = this.createChildId();
       this.setId(traceId);
       this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
-      const result = callable();
-      this.recordAnnotation(new Annotation.LocalOperationStop());
-      return result;
+      try {
+        return callable();
+      } catch (err) {
+        this.recordBinary('error', err.message ? err.message : err.toString());
+        throw err;
+      } finally {
+        this.recordAnnotation(new Annotation.LocalOperationStop());
+      }
     });
   }
 
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -8,6 +8,7 @@ const randomTraceId = require('./randomTraceId');
 const {now, hrtime} = require('../time');
 const {Endpoint} = require('../model');
 
+const isPromise = require('is-promise');
 
 function requiredArg(name) {
   throw new Error(`Tracer: Missing required argument ${name}.`);
@@ -81,19 +82,42 @@ class Tracer {
   }
 
   // creates a span, timing the given callable, adding any error as a tag
+  // if the callable returns a promise, a span stops after the promise resolves
   local(operationName, callable) {
+    if (typeof callable !== 'function') {
+      throw new Error('you must pass a function');
+    }
     return this.scoped(() => {
       const traceId = this.createChildId();
       this.setId(traceId);
       this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
+
+      let result;
       try {
-        return callable();
+        result = callable();
       } catch (err) {
         this.recordBinary('error', err.message ? err.message : err.toString());
-        throw err;
-      } finally {
         this.recordAnnotation(new Annotation.LocalOperationStop());
+        throw err;
       }
+
+      // Finish the span on a synchronous success
+      if (!isPromise(result)) {
+        this.recordAnnotation(new Annotation.LocalOperationStop());
+        return result;
+      }
+
+      // Ensure the span representing the promise completes
+      return result
+        .then((output) => {
+          this.recordAnnotation(new Annotation.LocalOperationStop());
+          return output;
+        })
+        .catch((err) => {
+          this.recordBinary('error', err.message ? err.message : err.toString());
+          this.recordAnnotation(new Annotation.LocalOperationStop());
+          throw err;
+        });
     });
   }
 
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -80,6 +80,17 @@ class Tracer {
     return childId;
   }
 
+  local(operationName, callable) {
+    return this.scoped(() => {
+      const traceId = this.createChildId();
+      this.setId(traceId);
+      this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
+      const result = callable();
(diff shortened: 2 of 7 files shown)


adriancole:
cc @narayaruna @fedj

packages/zipkin/src/tracer/index.js
```diff
@@ -69,6 +69,17 @@ class Tracer {
     return childId;
   }
 
+  local(operationName, callable) {
+    return this.scoped(() => {
+      const traceId = this.createChildId();
+      this.setId(traceId);
+      this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
+      const result = callable();
```
DanielMSchmidt: What about when the capable is a) no callable? b) throws an error?

eirslett: I guess we can assume that it's a callable, what else should it be?
But it could potentially return a Promise, in which case we should not annotate LocalOperationStop until the Promise is finished. (Async function with try-catch-finally could work, with `const result = await callable();`)

adriancole: thanks for the tips! I added a commit. I do need some advice on how to handle not-callable cc also @pbadenski as we need whatever we do to make sense in typescript (not sure you can overload functions in ts or not)

packages/zipkin/src/tracer/index.js
```diff
@@ -69,6 +69,23 @@ class Tracer {
...
   }
 
+  // creates a span, timing the given callable, adding any error as a tag
+  local(operationName, callable) {
+    return this.scoped(() => {
+      const traceId = this.createChildId();
+      this.setId(traceId);
+      this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
+      try {
+        return callable();
```
DanielMSchmidt: Seems way more solid, good job. For the promise return, we could simply add a check if it's a promise and chain a finally

packages/zipkin/src/tracer/index.js
```diff
@@ -69,6 +69,23 @@ class Tracer {
     return childId;
   }
 
+  // creates a span, timing the given callable, adding any error as a tag
+  local(operationName, callable) {
```
DanielMSchmidt: For the check we could either be explicit and throw an error directly, but this might be too disruptive for our users. I would go for a check if it's a function and if not simply return the callable

adriancole:
Thanks for the tips. Any favorite check logic or should I google?

On 11 Nov 2017 14:45, "Daniel Schmidt" <notifications@github.com> wrote:

*@DanielMSchmidt* commented on this pull request.
------------------------------

In packages/zipkin/src/tracer/index.js
<https://github.com/openzipkin/zipkin-js/pull/156#discussion_r150376519>:

> @@ -69,6 +69,23 @@ class Tracer {
     return childId;
   }

+  // creates a span, timing the given callable, adding any error as a tag
+  local(operationName, callable) {
+    return this.scoped(() => {
+      const traceId = this.createChildId();
+      this.setId(traceId);
+      this.recordAnnotation(new Annotation.LocalOperationStart(operationName));
+      try {
+        return callable();

Seems way more solid, good job. For the promise return, we could simply add
a check if it's a promise and chain a finally
------------------------------

In packages/zipkin/src/tracer/index.js
<https://github.com/openzipkin/zipkin-js/pull/156#discussion_r150376545>:

> @@ -69,6 +69,23 @@ class Tracer {
     return childId;
   }

+  // creates a span, timing the given callable, adding any error as a tag
+  local(operationName, callable) {

For the check we could either be explicit and throw an error directly, but
this might be too disruptive for our users. I would go for a check if it's
a function and if not simply return the callable

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/zipkin-js/pull/156#pullrequestreview-75925430>,
or mute the thread

DanielMSchmidt:
I would go for typeof callable === "function"

Adrian Cole <notifications@github.com> schrieb am Sa., 11. Nov. 2017, 12:45:

> Thanks for the tips. Any favorite check logic or should I google?
>

eirslett:
```javascript
function delay (ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
```
This is a simple JS function to create a Promise that resolves after the ms timeout.
We should have a unit test with a local trace that sleeps for some time (maybe 10 ms) and verify that the stop annotation is recorded after the sleep operation is done?

packages/zipkin/src/tracer/index.js
pbadenski: function overloading works in typescript, very natural too:
```
local<V>(name: string, callback: () => V): V;
local<V>(name: string, promise: Promise<V>): Promise<V>;
```

eirslett:
```
local<V>(name: string, promise: Promise<V>): Promise<V>;
```
No no no, you cannot pass in a promise like that, because that would mean you would have to start running your work before we have added the "before" annotations in Zipkin. It has to be

```
local<V>(name: string, callback: () => Promise<V>): Promise<V>;
```

then zipkin-js can decide when to start the work (typically immediately after adding its "before" annotations).

There is really no point of handling whether or not `callback` is callable or not, there is no other input than a callback that makes sense. So if we get something that is not callable, we should just throw a TypeError, I think.

packages/zipkin/src/tracer/index.js
eirslett: It will have to be more complex than that. Here are 5 unit test cases:

1) synchronous call, no error -> annotate, return success value
2) synchronous call, with error -> annotate, but rethrow the error
3) async/promise call, but throws a synchronous error -> annotate, but rethrow the error
4) async/promise call, but with async error (in a `.catch()` clause) -> annotate, but still rethrow the error so it's a rejected Promise
5) async/promise call, no error -> return the Promise with the success value

There will have to be some funky if-else-try-catch-ing going on...

adriancole:
I read that promise starts as soon as it is constructed. If so, we start
the span as soon as we wrap the promise, then add callback for on success
and failure?

adriancole:
Also seems promise is light on requirements. Is checking for "then" the
only way to tell?

eirslett:
```javascript
npm install is-promise
const isPromise = require('is-promise')
isPromise(result)
```

eirslett:
> I read that promise starts as soon as it is constructed.

Yes they do, so we won't even have time to wrap the Promise - it will already have started. Which is why we need a callback promise to construct the Promise once we have set up the "before" annotations.

adriancole:
ok I think it works as expected now. thanks @sundarsy for pairing on this. let us know if there's something that should be changed

packages/zipkin/index.d.ts
```diff
@@ -19,6 +19,7 @@ declare namespace zipkin {
     id: TraceId;
 
     scoped<V>(callback: () => V): V;
+    local<V>(name: string, callback: () => V): V;
```
adriancole: @pbadenski what should we put here to represent a callback that returns a `Promise<V>`? would it be `local<V>(name: string, callback: () => Promise<V>): V;` if so, is there an import needed? or do we declare it locally..

adriancole:
@voldern ptal if you don't mind

adriancole:
thanks for the review, folks. please comment here as I have a couple things I'd like an opinion on before merge https://github.com/openzipkin/zipkin-js/issues/104#issuecomment-345413722

eirslett:
LGTM!

adriancole:
PTAL at https://github.com/openzipkin/zipkin-js/pull/159 which we'll need (or some variant of) to ensure the local span names can be used in search

adriancole:
ok ready to merge.. last warning :)

adriancole:
promise tracing was actually broke.. this fixes it. can someone do a quick review please? https://github.com/openzipkin/zipkin-js/pull/165

PR #158, opened 2017-11-17
Honors sampling flag by not recording data when false

### LINKED ISSUE #152: expressMiddleware.js does not check for the sampled flag
I've been having trouble getting sampling and B3 propagation, specifically with the "X-B3-Sampled" header to work with the zipkin-instrumentation-express package.  It seems that regardless of what the sampler is set to or whether or not the incoming request has the X-B3-Sampled header set to 0 the request is logged.  After debugging through the runtime it seems like the the "sampled" flag is being set properly by the library, but the expressMiddleware simply never looks at it.  As a workaround I've wrapped the "recordResponse" methods in an if statement like so:

```
      var id = instrumentation.recordRequest(req.method, formatRequestUrl(req), readHeader);
      if (id.sampled.value) {
        res.on('finish', function () {
          tracer.scoped(function () {
            instrumentation.recordResponse(id, res.statusCode);
          });
        });
      }
```
file in question: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-express/src/expressMiddleware.js

I wanted to know if this was the intended behavior or if there is a check that happens in another package.  Looking at the zipkin libraries in some other languages there are more obvious checks.

adriancole:
Before, we were unconditionally recording annotations and tags.

Fixes #152

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -86,11 +86,15 @@ class Tracer {
   }
 
   recordAnnotation(annotation) {
-    this.recorder.record(new Record({
-      traceId: this.id,
-      timestamp: now(this._startTimestamp, this._startTick),
-      annotation
-    }));
+    this.id.sampled.ifPresent(sampled => {
+      // sampled present is different than sampled == true
+      if (!sampled) return;
+      this.recorder.record(new Record({
+        traceId: this.id,
+        timestamp: now(this._startTimestamp, this._startTick),
+        annotation
+      }));
+    });
   }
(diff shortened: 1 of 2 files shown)


adriancole:
So what's happening before is that when a trace is started, regardless of whether it is sampled or not, it is reported to zipkin. This is counter-intuitive, as most would expect a 1-1 relationship between sampling status and whether data goes to zipkin.

In the future, we might have a separate local cache for data (even if it doesn't go to zipkin), but for now, the most sensible choice is to only record data when sampled=true

PR #159, opened 2017-11-19
Centralizes assignment of local service name

adriancole:
This adds `localServiceName` and `localEndpoint` constructor parameters to `Tracer`, which avoids repeating the same for all instrumentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,40 @@ how tracing services looks.
 
 `npm install zipkin --save`
 
+## Basic Setup:
+
+```javascript
+const {
+  Tracer,
+  BatchRecorder,
+  jsonEncoder: {JSON_V2}
+} = require('zipkin');
+const CLSContext = require('zipkin-context-cls');
+const {HttpLogger} = require('zipkin-transport-http');
+
+// Setup the tracer to use http and implicit trace context
+const tracer = new Tracer({
+  ctxImpl: new CLSContext('zipkin'),
+  recorder: new BatchRecorder({
+    logger: new HttpLogger({
+      endpoint: 'http://localhost:9411/api/v2/spans',
+      jsonEncoder: JSON_V2
+    })
+  }),
+  localServiceName: 'service-a' // name of this application
+});
+
+// now use tracer to construct instrumentation! For example, fetch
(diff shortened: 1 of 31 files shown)


README.md
```diff
@@ -13,11 +13,40 @@ how tracing services looks.
...
+    })
+  }),
+  localServiceName: 'service-a' // name of this application
+});
+
+// now use tracer to construct instrumentation! For example, fetch
+const wrapFetch = require('zipkin-instrumentation-fetch');
+
+const remoteServiceName = 'youtube';
+const zipkinFetch = wrapFetch(fetch, {tracer, remoteServiceName});
```
adriancole: @yonran @pbadenski hoping we can somehow deprecate passing the `serviceName` argument into each instrumentation. People tend to get it wrong which results in traces with mixed service names for the same service, or broken service dependency graphs. By moving this to the tracer, hopefully it will be more obvious.

Do you know how to deprecate something in javascript?

Also, I'd like to make sure we don't encourage passing `serviceName` in typescript. Is the best way to just not define the arg?

adriancole:
rebased over master (which fixes appveyor)

packages/zipkin/index.d.ts
```diff
@@ -30,7 +30,7 @@ declare namespace zipkin {
   }
 
   class Tracer {
-    constructor(args: { ctxImpl: Context<TraceId>, recorder: Recorder, sampler?: sampler.Sampler, traceId128Bit?: boolean });
+    constructor(args: { ctxImpl: Context<TraceId>, recorder: Recorder, sampler?: sampler.Sampler, traceId128Bit?: boolean, localServiceName?: string, localEndpoint?: model.Endpoint });
```
adriancole: @yonran is it correct typscript to do `model.Endpoint` as I am here (to reference a type under a different namespace in the same file)?

README.md
yonran: That’s fine that you took it out of the TypeScript declaration.

packages/zipkin/index.d.ts
yonran: Yes, tsc is happy (`node_modules/.bin/tsc --strict index.d.ts`), and Visual Studio Code can jump to the correct declaration.

PR #160, opened 2017-11-19
Typescript declarations for zipkin-instrumentation-express and zipkin-instrumentation-request

yonran:
Fix `packages/zipkin/index.d.ts`, which had some duplicate declarations because pawel’s commit dc62adeb3dc47eafff02abedff50356ba5f8278f was rebased on top of my commits 95abee9f43165cd30e9066567df260e66a8e1ee1, e387cfa9ea4c7c14b5d1650c062ed253818e54b0, fedc9bc2ed10dad9de429df5b5cef84d89324a70.

Add typescript declarations for zipkin-instrumentation-express and zipkin-instrumentation-request

### DIFF
--- a/packages/zipkin-instrumentation-express/index.d.ts
+++ b/packages/zipkin-instrumentation-express/index.d.ts
@@ -1,0 +1,22 @@
+import {Tracer} from "zipkin"
+import {Handler} from "express"
+
+/**
+ * When a request comes in, creates ServerRecv annotation and then passes
+ * the request to the next middlewhare. When the final middleware finishes
+ * the request, creates ServerSend annotation
+ *
+ * Sets the tracer.id to the span id prior to running the next middleware.
+ *
+ * Note that if the next middleware makes async calls, it should either
+ * store the span id manually or use a CLSContext so that the annotations
+ * go to the correct spans
+ */
+export declare function expressMiddleware(
+  options: {tracer: Tracer, serviceName?: string, port?: number}
+): Handler;
+
+export declare function wrapExpressHttpProxy(
+  proxy: (host: string, options?: any) => Handler,
+  options: {tracer: Tracer, serviceName?: string, remoteServiceName?: string}
+): (host: string, options?: any) => Handler
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -17,10 +17,12 @@ declare namespace zipkin {
   namespace sampler {
     class Sampler {
       constructor(evaluator: (traceId: TraceId) => boolean);
+      shouldSample(traceId: TraceId): option.IOption<boolean>
     }
 
     class CountingSampler implements Sampler {
       constructor(sampleRate?: number);
+      shouldSample(traceId: TraceId): option.IOption<boolean>
     }
 
     const neverSample: (traceId: TraceId) => boolean;
@@ -54,11 +56,11 @@ declare namespace zipkin {
     _sampled:  option.IOption<boolean>;
     _flags:    number;
 
-    traceId:  string;
-    parentId: string;
-    spanId:   string;
-    sampled:  boolean;
-    flags:    number;
+    readonly traceId: string;
+    readonly parentId: string;
+    readonly spanId: string;
+    readonly sampled: option.IOption<boolean>;
+    readonly flags: number;
 
     isDebug(): boolean;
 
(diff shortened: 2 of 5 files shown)


packages/zipkin-instrumentation-express/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+ * the request, creates ServerSend annotation
+ *
+ * Sets the tracer.id to the span id prior to running the next middleware.
+ *
+ * Note that if the next middleware makes async calls, it should either
+ * store the span id manually or use a CLSContext so that the annotations
+ * go to the correct spans
+ */
+export declare function expressMiddleware(
+  options: {tracer: Tracer, serviceName?: string, port?: number}
```
adriancole: made a note about possibly removing serviceName from here as the `localServiceName` and `localEndpoint` arg for Tracer ends in less repetition.

packages/zipkin-instrumentation-express/index.d.ts
```diff
@@ -0,0 +1,22 @@
+import {Tracer} from "zipkin"
+import {Handler} from "express"
+
+/**
+ * When a request comes in, creates ServerRecv annotation and then passes
```
adriancole: what's the usual strategy for docs? put them in typescript, then make a note in the source javascript to look at the ts?

packages/zipkin/index.d.ts
```diff
@@ -272,20 +268,8 @@ declare namespace zipkin {
     letContext<V>(ctx: TraceId, callback: () => V): V;
   }
 
-  namespace sampler {
```
adriancole: dupe huh?

packages/zipkin/index.d.ts
```diff
@@ -272,20 +268,8 @@ declare namespace zipkin {
...
-    }
-    function neverSample(traceId: TraceId): boolean
-    function alwaysSample(traceId: TraceId): boolean
-    class CountingSampler extends Sampler {
-      constructor(sampleRate: number)
-    }
-  }
-
-  class Request {
+  namespace Request {
```
adriancole: make a comment why namespace vs class?

packages/zipkin/index.d.ts
```diff
@@ -120,7 +116,7 @@ declare namespace zipkin {
   }
 
   namespace model {
-    interface Endpoint {
+    class Endpoint {
```
adriancole: note why class not interface? (this is me curious not saying it should be interface)

adriancole:
ps if you rebase off master, appveyor should be fixed

packages/zipkin-instrumentation-express/index.d.ts
yonran: I don’t know what people normally do. If you want I can move it to the js file

packages/zipkin-instrumentation-express/index.d.ts
yonran: Should I make a note?

packages/zipkin/index.d.ts
yonran: If it’s an interface, then you can’t say `new Endpoint`, and the `constructor` is incorrectly interpreted as a method returning `any`.

packages/zipkin/index.d.ts
yonran: Yes I added a declaration for `Sampler` in 95abee9f43165cd30e9066567df260e66a8e1ee1, and pawel’s commit dc62adeb3dc47eafff02abedff50356ba5f8278f, which was rebased on top, also contained `Sampler`.

packages/zipkin/index.d.ts
yonran: If there is a `constructor`, it should be a `class`.

packages/zipkin-instrumentation-express/index.d.ts
adriancole: this is fine for now.. thx

packages/zipkin-instrumentation-express/index.d.ts
adriancole: maybe take a look at https://github.com/openzipkin/zipkin-js/pull/159#discussion_r151887803? I am just wondering if we should document the serviceName param as it probably should be gone

adriancole:
thx for the tips. all looks good as is.

PR #164, opened 2017-11-22
add support for optional fetch timeout

### LINKED ISSUE #163: Add support for timeout option to HttpLogger
Hi! Was thinking it would be nice to be able to initialize the `HttpLogger` with arbitrary fetch options, but specifically, we'd like to be able to provide a [timeout value for `node-fetch`](https://github.com/bitinn/node-fetch#fetch-options) to abort the underlying request in the event that the Zipkin endpoint does not respond quickly enough for the application..

```javascript
// if just the timeout
let logger = new HttpLogger({ endpoint: 'http://endpoint.zipkin' , /* ... */, timeout: 2000 });

// if arbitrary
let logger = new HttpLogger({
  endpoint: 'http://endpoint.zipkin',
  /* ... */, 
  fetchOptions: { timeout: 2000 },
});
```

the supply them to the fetch client at

https://github.com/openzipkin/zipkin-js/blob/3a24206e004a774735b5a5e9ed5648e945890ac4/packages/zipkin-transport-http/src/HttpLogger.js#L38-L42

Seems straightforward, but I wanted to check with the project before starting on the implementation. Any concerns?

montmanu:
had some trouble figuring out how to test this.. any suggestions? I started down the path of something like:

```javascript
it('should abort the request if response exceeds timeout', function(done) {
  // ..
  app.post('/api/v1/spans', (req, res) => {
    setTimeout(() => { res.status(202).json({}); }, 61000);
  });
  // ...
  this.server = app.listen(0, () => {
    // ..
    const httpLogger = new HttpLogger({
      endpoint: `http://localhost:${this.port}/api/v1/spans`,
      headers: {Authorization: 'Token'},
      jsonEncoder: JSON_V2,
      timeout: 2000,
    });
    // .. 
  });
});
```

But I'm not sure how to get a reference to the promise returned by fetch.. any advice would be welcome!

closes #163

### DIFF
--- a/packages/zipkin-transport-http/index.d.ts
+++ b/packages/zipkin-transport-http/index.d.ts
@@ -1,7 +1,7 @@
 import {JsonEncoder, Logger, model} from "zipkin"
 
 declare class HttpLogger implements Logger {
-  constructor(options: {endpoint: string, httpInterval?: number, jsonEncoder?: JsonEncoder});
+  constructor(options: {endpoint: string, httpInterval?: number, jsonEncoder?: JsonEncoder, httpTimeout?: number});
   logSpan(span: model.Span): void;
 }
 export {HttpLogger}
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -11,7 +11,7 @@ const {
 } = require('zipkin');
 
 class HttpLogger {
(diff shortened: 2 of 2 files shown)


adriancole:
not sure how to make the test.. usually I would use a test server which has a delay function. I'll trust you to follow-up if someone has advice. Thanks!

adriancole:
0.11.2

PR #165, opened 2017-11-22
Closes the correct span on promise complete

adriancole:
Fixes a bad bug if you actually used the local tracing with a promise. I noticed found this using the react-native example. Immediate patch release to follow merge.

### DIFF
--- a/packages/zipkin/test/trace.test.js
+++ b/packages/zipkin/test/trace.test.js
@@ -115,9 +115,7 @@ describe('Tracer', () => {
     ctxImpl.scoped(() => {
       const promise = trace.local('buy-smoothie', () =>
         Promise.delay(10)
-               .then(() => {
-                 throw new Error('no smoothies. try our cake');
-               })
+               .then(() => 'smoothie')
       );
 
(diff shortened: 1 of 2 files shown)


adriancole:
going to merge and cut this as currently things are broke. welcome feedback after commit (this change is not api affecting, so safe to suggest rework!)

packages/zipkin/test/trace.test.js
```diff
@@ -115,9 +115,7 @@ describe('Tracer', () => {
     ctxImpl.scoped(() => {
       const promise = trace.local('buy-smoothie', () =>
         Promise.delay(10)
-               .then(() => {
-                 throw new Error('no smoothies. try our cake');
-               })
+               .then(() => 'smoothie')
```
eirslett: Shouldn't we have a unit test for the error case as well?

adriancole: :) the red you are seein was me correcting code that did same check in 2 test cases! I fixed one to test success

adriancole: Can add another test to verify the error calback sets correct trace ids though..

PR #166, opened 2017-11-26
added error handling for http-transport

Celadora:
adds error handling to zipkin-transport-http

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -10,11 +10,16 @@ const {
   jsonEncoder: {JSON_V1}
 } = require('zipkin');
 
-class HttpLogger {
+const EventEmitter = require('events').EventEmitter;
+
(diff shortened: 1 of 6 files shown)


packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -26,28 +31,44 @@ class HttpLogger {
...
+      fetch(self.endpoint, {
         method: 'POST',
         body: postBody,
-        headers: this.headers,
+        headers: self.headers,
       }).then((response) => {
         if (response.status !== 202) {
-          console.error('Unexpected response while sending Zipkin data, status:' +
-            `${response.status}, body: ${postBody}`);
+          var err = 'Unexpected response while sending Zipkin data, status:' +
```
adriancole: I might move the string cat to the handler.. presumably there is some cost to this

adriancole:
I imagine there are some linting problems that need to be sorted

adriancole:
thanks for the change. I am linting it now

PR #168, opened 2017-11-29
Fix null reference error in BatchRecorder when terminating a span tha…

spoulson:
…t has been expired.

### DIFF
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -40,6 +40,12 @@ class BatchRecorder {
 
   _writeSpan(id) {
     const span = this.partialSpans.get(id);
+
+    if (span === undefined) {
+      // Span not found.  Could have been expired.
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


spoulson:
AppVeyor build fails on unrelated unit test.  They all pass for me locally.

eirslett:
I don't like the architecture of the existing code... this is a patch that fixes the problem, but I'm not sure _writeSpan() should handle things like non-existing spans... it's a violation of the single responsibility principle. @adriancole do you think it's possible to make some more invasive refactoring in this file?

adriancole:
@eirslett yep certainly we can. agreed that writing the span should be on something that is in fact a span at the time :)

rooftopsparrow:
This is affecting me as well, is there a previous version that isn't affected by this ?

adriancole:
I added a TODO to follow-up on the refactor rather than block the fix

adriancole:
0.11.2

spoulson:
🎉

PR #170, opened 2017-12-05
TypeScript type fix, and add fields

yonran:
Fix the type of ipv4 for Endpoint constructor

Add fields of the Annotation subclasses


adriancole:
looks right to me. thanks!

Issue #171, opened 2017-12-05
Issue while propagating traceId due to stringToBoolean function


frankcs:
Hi there, thanks for all the work put into this.

I'm having an small issue when trying to propagate the trace id received from a service implemented in `go` using openzipkin/zipkin-go-opentracing for tracing and traces are not being properly recorded in zipkin. I'm using zipkin-instrumentation-express for my express app.

I tracked the problem down to this file:
`zipkin-js/packages/zipkin/src/instrumentation/httpServer.js`

My issue is that the `sampled: sampled.map(stringToBoolean)` flag is being incorrectly set as the parameter received by:

```
function stringToBoolean(str) {
  return str === '1';
}
```
is always `'true'`

I fixed it locally by doing this:

```
function stringToBoolean(str) {
  return str === '1' || str === 'true';
}
```
And now all is properly being recorded in zipkin. 

Is this something we should fix here or is the data being received wrong?

adriancole:
yeah B3 is 1 or 0 for sampled bit, although some old libraries (that
predate zipkin-go-opentracing) sent 'true' or 'false' literals. We can
tolerantly parse the 'true' 'false' but should certainly only write '1' or
'0'

https://github.com/openzipkin/b3-propagation

frankcs:
Created a PR for it
https://github.com/openzipkin/zipkin-js/pull/172

PR #172, opened 2017-12-05
Tolerating boolean literals sent by libraries like zipkin-go-opentracing in b3

frankcs:
https://github.com/openzipkin/zipkin-js/issues/171


adriancole:
probably should add a unit test so this stays fixed?

frankcs:
I could use some help here with the test, I'm not sure if this is the best way check for that.

adriancole:
what you did works fine. Thanks!

Issue #176, opened 2018-01-04
Wrong traceID between contiguous requests


mkante:
Recently we have been trying to instrument our NodeJs microservices with zipkin, unfortunatly we are running into a strange issue with **Memcached** callbacks. The memcached callbacks shows the previous request traceIDs if the requests are contiguous (few seconds apparts).

To help properly explain my bug I setup a small microservice with zipkin, zipkin-instrumentation-memcached, zipkin-instrumentation-express. You will need a running memcached server to follow the steps.

Im using `node v6.11.1` with `npm v3.10.10`

Here is my folder structure

```
/zipkin-test
  |- /node_modules
  |- package.json
  |- service.js
```

Inside **package.js**

```
{
  "name": "zipkin-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "async": "^2.6.0",
    "body-parser": "^1.18.2",
    "express": "^4.16.2",
    "memcached": "^2.2.2",
    "request": "^2.83.0",
    "request-promise": "^4.2.2",
    "zipkin": "^0.11.1",
    "zipkin-context-cls": "^0.11.0",
    "zipkin-instrumentation-express": "^0.11.1",
    "zipkin-instrumentation-memcached": "^0.11.1",
    "zipkin-instrumentation-request": "^0.11.1"
  }
}
```

Inside **service.js**

```
var Tracer = require('zipkin').Tracer;
var ConsoleRecorder = require('zipkin').ConsoleRecorder;
var CLSContext = require('zipkin-context-cls');
var request = require('request-promise');
var app = require('express')();
var wrapRequest = require('zipkin-instrumentation-request');
var ZipkinClient = require('zipkin-instrumentation-memcached');
var Memcached =  require('memcached')
var async = require('async');
var zipkinMiddleware = require('zipkin-instrumentation-express').expressMiddleware;

var log = console.log;

// Create a tracer
var tracer = new Tracer({
  ctxImpl: new CLSContext('zipkin'),
  traceId128Bit: true,
  recorder: new ConsoleRecorder(function(){}),
  localServiceName: 'service-3'
});

log('Root TraceID: ', tracer.id.traceId);

var zipkinRequest = wrapRequest(request, { tracer: tracer });
var MemClient = new ZipkinClient(tracer, Memcached);
[...]

PR #177, opened 2018-01-06
ticket-176

mkante:
Fix traceID restore problem caused by CLS


eirslett:
@adriancole do you have time to look at this issue/PR? I'm not 100 % sure it's the correct solution to the problem, but at least it looks like an improvement! I'm a bit concerned it could cause regressions or changed behavior in other cases. But unfortunately, I don't have time to look closely at it myself.

mkante:
I see `continuous-integration/appveyor/pr` failure from Kafka,  I don't think it's related to my PR. Once  have an environment setup with kafka, I will look into it.

adriancole:
thank you. I will make time to look at this, as it is important. Thanks for the work here @mkante!

packages/zipkin-instrumentation-memcached/src/zipkinClient.js
```diff
@@ -17,6 +17,7 @@ module.exports = function zipkinClient(
         }));
         tracer.recordAnnotation(new Annotation.ClientRecv());
       });
+      tracer.setId(originalID);
```
adriancole: so this line implies a leak I guess

mkante: Correct

mkante: I shared small service on [issue 172](https://github.com/openzipkin/zipkin-js/issues/176) to prove my point

adriancole: visual check, all calls to setId happen in scoped blocks, so this guard is indeed interesting

adriancole: the test passes with this commented out.. however the test uses explicit context, not CLS. CLS is the scenario of your example I think...

adriancole: this also passes with CLS..

adriancole:
after #178 I think the below patch might make things slightly more explicit. Only thing I have concern with is that since the test breaks before and after.. we should write a test that breaks before :P are you able to help with that?

```diff
-  function mkZipkinCallback(callback, id, originalID) {
+  function mkZipkinCallback(callback, id) {
+    const originalId = tracer.id;
     return function zipkinCallback(...args) {
       tracer.letId(id, () => {
         // TODO: parse host and port from details in callback
@@ -16,8 +17,10 @@ module.exports = function zipkinClient(
         }));
         tracer.recordAnnotation(new Annotation.ClientRecv());
       });
-      tracer.setId(originalID);
-      callback.apply(this, args);
+      // callback runs after the client request, so let's restore the former ID
+      tracer.letId(originalId, () => {
+        callback.apply(this, args);
+      });
     };
   }
```

adriancole:
meant to say the test passes before and after the change to src/

adriancole:
probably to get the test to fail, we need to manually manipulate the trace ID in one of the callbacks or somehow get the test to run twice

mkante:
If I understand you want to merge #178 first, then I can apply the code above? That wont be problem. Do you know if #178 is going in any time soon?

adriancole:
178 just needs a review.. care to?

mkante:
No problem

PR #178, opened 2018-01-08
Trades unused letChildId for letId

adriancole:
This adds letId which is a shortcut to combining tracer.scoped with
setId. It pays for the additional code by removing an unused function:
tracer.letChildId


adriancole:
versions will be sorted on release. `yarn test` does the right things in the mean time afaik

adriancole:
we have a persistent issue in the node 4 tests which is behind the red

mkante:
Cool thx

adriancole:
@pbadenski do you have time to update the http instrumentation to use letId? I think it might be less errorprone, but could be wrong

PR #180, opened 2018-01-22
Add types to hapi instrumentation

kevin-greene-ck:


adriancole:
I don't think we should encourage `serviceName` anymore as it muddies Tracer's property `localServiceName`, which it should be equivalent to. For example, hapi's readme is correct.

reverting the `serviceName` part, sgtm!

kevin-greene-ck:
@adriancole Ah, I see. However, the types should still report the property as long as the underlying JS does, I would think. Making the property optional.

adriancole:
lack of publishing this property in ts is the ultimate way to discourage usage, and conversely, if we publish it into ts we encourage its use (even if optional).

Please don't add serviceName to typescript as it encourages its use, which is deprecated and should be removed.

adriancole:
thanks @kevin-greene-ck!

PR #188, opened 2018-02-07
Fix memory leak for not closing kafka producer listeners

jemapi:
Remove listeners when producer sends or fails to send, to avoid memory leaks when trace has a lot of spans.


adriancole:
thanks very much

PR #193, opened 2018-03-02
Fix return type of HttpServer.recordRequest

kevin-greene-ck:
Instrumentation.HttpServer.recordRequest returns a TraceId object, not a string.
Instrumentation.HttpClient/HttpClient.recordResponse takes a TraceId object, not a string.


adriancole:
I agree! thanks

Issue #183, opened 2018-01-30
Use a 


jcchavezs:
Currently `hapi` middleware uses the `http method` for naming spans. This solution is a not so bad low cardinality option, however from user perspective this could be improved by using the `path` instead of the `verb`. For that we can override the span name in the `onPreResponse` (see [this](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-hapi/src/hapiMiddleware.js#L41)) event so instead of having `get` as span name we get `get /user/{user_id}`.

@fbaiodias

adriancole:
There is some discussion about "http.template" which sounds like this
change can afford.

Not sure if we should change default because it is what most zipkin
instrumentation use since the beginning (ex will mix up drop down), but
should be possible (ex via a hook or a naming policy config)

So regardless we should support the mechanism just not sure if it should be
the default policy or not (as it isnt universally or even mostly supported
across frameworks it will end up inconsistent)

jcchavezs:
A test suite for http has been added and now to test it works we just need to set this to true: https://github.com/openzipkin/zipkin-js/blob/a173522b73c71f2ac330b7b2d41959c50d19762a/test/httpServerTestFixture.js#L13

PR #200, opened 2018-03-27
Add route path to express middleware

epallerols:
Adds the route path to the RPC annotation.

Tests are failing, it looks like an issue on the tests. The req.route is not yet available when running the test. We have tested it in one of our repos and is always defined

similar to this issue but for express https://github.com/openzipkin/zipkin-js/issues/183

ping @jcchavezs

### DIFF
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -27,8 +27,9 @@ module.exports = function expressMiddleware({tracer, serviceName, port = 0}) {
         }
       }
 
+      const name = req.method + ' ' + req.route.path;
       const id =
-        instrumentation.recordRequest(req.method, formatRequestUrl(req), readHeader);
+        instrumentation.recordRequest(name, formatRequestUrl(req), readHeader);
 
       res.on('finish', () => {
         tracer.scoped(() => {


jcchavezs:
Ping @adriancole

adriancole:
by not yet, you mean if you move the code later it finds it?

adriancole:
maybe look at this for a more complete example including tests? https://github.com/openzipkin/zipkin-js/commit/4e8fdcccc5a5fba08044d070fb99ef6de9de93be

jcchavezs:
@epallerols I believe this was covered in https://github.com/openzipkin/zipkin-js/pull/226/files?

Ping @pchiwan

adriancole:
I think you can overwrite the RPC name annotation. not sure you know that..

jcchavezs:
I remember we added a third parameter to `instrumentation.recordResponse(id,
res.statusCode)` to receive a name as optional third parameter. Not sure
how good was that. Other than that you don't have the name yet.

Den fre. 30. nov. 2018, 23:59 skrev Adrian Cole <notifications@github.com:

adriancole:
in brave we literally had to make a special parse case for data from
response stage... this was one of the reasons.

ghermeto:
> I remember we added a third parameter to `instrumentation.recordResponse(id, res.statusCode)` to receive a name as optional third parameter. Not sure how good was that. Other than that you don't have the name yet.

@jcchavezs I'm not sure if I follow... you meant the third parameter on `recordResponse` on which file? `httpServer` 3rd param is an error: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/instrumentation/httpServer.js#L90

if we alter the middleware to look like this will work:

```javascript
      res.on('finish', () => {
        tracer.scoped(() => {
          tracer.setId(id);
          tracer.recordRpc(`${req.method} ${req.route.path}`);
          instrumentation.recordResponse(id, res.statusCode);
        });
      });
```
if there is interest I can send a PR with this change and proper tests.

While we are at it, when do we really want to record requests? Only on `finish` or on `finish OR close` (like Restify)?  If the [request is aborted by the server](https://github.com/nodejs/node/blob/39d25eb1e2c6ffbb02b28658a1307d047370af39/lib/_http_server.js#L443-L451), do we want the span?

jcchavezs:
> While we are at it, when do we really want to record requests? Only on
finish or on finish OR close (like Restify)? If the request is aborted by
the server
<https://github.com/nodejs/node/blob/39d25eb1e2c6ffbb02b28658a1307d047370af39/lib/_http_server.js#L443-L451>,
do we want the span?

Reporters usually send spans only when the span is finished:
https://github.com/openzipkin/zipkin-go/blob/master/span_implementation.go#L69
cc @adriancole

Den lør. 1. des. 2018, 02:06 skrev Guilherme Hermeto <
notifications@github.com:

jcchavezs:
Closed by https://github.com/openzipkin/zipkin-js/pull/320

PR #203, opened 2018-04-03
Remove chai from transport-http dependencies

### LINKED ISSUE #202: zipkin-transport-http claims chai as a dependency in 0.12.0
Pull request inbound to remove this addition and rely on the top level package's devDependencies.

rooftopsparrow:
Fixes #202


adriancole:
thanks!

PhiLhoSoft:
I came to report about this unneeded dependency, but I see I should update instead, thanks.
BTW, I feel zipkin should be moved from devDependencies to dependencies.

PR #206, opened 2018-04-15
Adds proxy argument for routing HttpLogger through a proxy

### LINKED ISSUE #205: Support zipkin server behind a proxy in HttpLogger
**Problem**
We have two networks (let A and B) that can only communicate with each other through a proxy gateway. When specifying the address to the Zipkin server in network B from an application in network A, the HttpLogger is attempting to connect directly without going through the proxy, resulting in a failure to post the trace.

Problem experienced using version `0.12.0` of the `zipkin-transport-http` package using Node v8.10.0 for Ubuntu 16.04.

**Solution**
I am working on implementing the solution outlined in the `node-fetch` repository under [node-fetch Options](https://github.com/bitinn/node-fetch#options) to be able to specify a `http.agent` in the fetch parameter.

zephinzer:
Fixes #205

### DIFF
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -13,12 +13,15 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
+    "http-proxy-agent": "^2.1.0",
+    "https-proxy-agent": "^2.2.1",
     "node-fetch": "^1.5.3"
   },
   "devDependencies": {
     "babel-cli": "^6.23.0",
     "body-parser": "^1.15.2",
     "express": "^4.13.4",
+    "http-proxy-middleware": "^0.18.0",
     "mocha": "^3.2.0",
     "zipkin": "^0.12.0"
   }
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -2,18 +2,26 @@
 const globalFetch =
   (typeof window !== 'undefined' && window.fetch) ||
   (typeof global !== 'undefined' && global.fetch);
-
 // eslint-disable-next-line global-require
 const fetch = globalFetch || require.call(null, 'node-fetch');
 
 const {
   jsonEncoder: {JSON_V1}
 } = require('zipkin');
+const HttpsProxyAgent = require('https-proxy-agent');
+const HttpProxyAgent = require('http-proxy-agent');
 
 const EventEmitter = require('events').EventEmitter;
 
 class HttpLogger extends EventEmitter {
-  constructor({endpoint, headers = {}, httpInterval = 1000, jsonEncoder = JSON_V1, timeout = 0}) {
+  constructor({
+    endpoint,
+    headers = {},
+    httpInterval = 1000,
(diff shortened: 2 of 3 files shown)


packages/zipkin-transport-http/package.json
```diff
@@ -13,12 +13,15 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
+    "http-proxy-agent": "^2.1.0",
```
adriancole: what's the risk of having these dependencies?

zephinzer:
sorry, it's my first time hearing that question, how should i assess that?

adriancole:
On Mon, Apr 16, 2018 at 6:44 PM, Joseph Matthias Goh <
notifications@github.com> wrote:

> sorry, it's my first time hearing that question, how should i assess that?
>

So, I guess I'm assuming that this transport is used by a lot of users
users. Sometimes dependencies can conflict with application code,
especially if they change incompatibly a lot or pull a string of deps that
can also conflict. Sortof asking for some summary judgement

zephinzer:
i see, the `http-proxy-agent` and `https-proxy-agent` together adds 7 dependencies including itself. the five dependencies are `debug`, `ms`, `agent-base`, `es6-promise` and `es6-promisify` which are widely used libraries, i don't assess it is likely there will be such conflicts caused by irregular release cycles/improper versioning. also, the proxy is only invoked upon passing of the argument so it will not break existing implementations

adriancole:
do you anticipate this would break those using react or browser functionality? cc @DanielMSchmidt  @micaelsampaio

zephinzer:
@adriancole i suspect it might, didn't consider that- are there any known open source projects using browser functionality so i can quickly replicate it to test?

jcchavezs:
@zephinzer could you proceed with the test you mention?

adriancole:
lacking an automatic  browser test, it would in my opinion suffice if you manually test by locally publishing this, editing the zipkin-js-example to use a proxy, and testing it

https://github.com/openzipkin/zipkin-js-example

I don't actually know of an automatic browserify test, and how to stitch it into circleci... maybe @eirslett does perhaps, but seems a separate issue to create that infra cc @openzipkin/devops-tooling

packages/zipkin-transport-http/package.json
jcchavezs: Do you mean security, amount of dependencies or compatibility?

jcchavezs:
@zephinzer could you please rebase this? Also, any chance on continuing this?

jcchavezs:
@zephinzer could we rebase this?

jcchavezs:
@zephinzer could you please rebase this? Also, any chance on continuing this?

jcchavezs:
Any chance instead we add the dependencies, we just add an interface for proxy and allow users to inject it?

adriancole:
decoration somehow sounds good.

strategically: bear in mind we probably will want to redo the http logger internals soon, so that the "sender" concept is extracted. This will reduce the code about http to very small and makes proxy or alternate impls easy.

jcchavezs:
This is fixed in https://github.com/openzipkin/zipkin-js/pull/439 hence I am closing this issue. Thanks @zephinzer anyways for the effort!

PR #213, opened 2018-05-23
kafkaLogger unhandled error in promise fix

john-bui-zocdoc:
Removed promise in KafkaLogger.js and provide access to the producer within logSpan. onError is always bound to the producer (not sure why removeAllListeners is ever called), and can be defined in the constructor.


adriancole:
thanks you'll need to run `yarn lint` when ready for review

adriancole:
thanks for the help!

PR #219, opened 2018-06-13
Adds headers option to HttpLogger types

kevin-greene-ck:
The HttpLogger accepts an option of Http headers to add to the outgoing request. This adds support for that option to the accompanying type definition file.


adriancole:
thanks

Issue #220, opened 2018-06-15
Restify middleware not reporting errors


russell-matt:
The restify middleware instrumentation is currently not handling error status codes, and thus is not marking the span with the error annotation. This results in Zipkin UI frameworks to not correctly highlight the span as such.

PR #221, opened 2018-06-15
Addresses issue #220

russell-matt:
Added functionality to pass an error to the [httpServer](https://github.com/openzipkin/zipkin-js/blob/6a12a1d3bb0567e7e3e3e69a69a1c4bffdeb0351/packages/zipkin/src/instrumentation/httpServer.js#L86) recordResponse function on status codes >=400 to indicate them as an error.

### DIFF
--- a/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
+++ b/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
@@ -35,7 +35,7 @@ module.exports = function restifyMiddleware({tracer, serviceName, port = 0}) {
 
         let error = null;
 
-        if(res.statusCode >= 400) {
+        if (res.statusCode >= 400) {
           error = res.statusCode;
         }
 
--- a/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
+++ b/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
@@ -3,6 +3,8 @@ const {
   Instrumentation
 } = require('zipkin');
 const url = require('url');
+const LOWER_LIMIT = 200;
+const UPPER_LIMIT = 399;
 
 function headerOption(req, header) {
   const val = req.header(header);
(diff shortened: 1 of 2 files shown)


packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
```diff
@@ -33,8 +33,14 @@ module.exports = function restifyMiddleware({tracer, serviceName, port = 0}) {
...
 
+        const error = null;
+
+        if(res.statusCode >= 400) {
+          error = res.statusMessage;
+        }
+
         tracer.scoped(() => {
-          instrumentation.recordResponse(id, res.statusCode);
+          instrumentation.recordResponse(id, res.statusCode, error);
```
adriancole: fyi typically we set error to the same as the status code

packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
```diff
@@ -33,8 +33,14 @@ module.exports = function restifyMiddleware({tracer, serviceName, port = 0}) {
         res.removeListener('close', onCloseOrFinish);
         res.removeListener('finish', onCloseOrFinish);
 
+        let error = null;
```
eirslett: It's better to do this immutably;

```
const error = res.statusCode >= 400 ? res.statusCode : null;
```

And maybe also better to move it inside the `tracer.scoped(() => ` block (after what looks like line 42), so it's defined closer to where it's used?

eirslett:
@adriancole Is this behaviour consistent with the other tracers, like brave? Many people have many different opinions about treating HTTP status codes as errors, or not...

adriancole:
in brave the default http "error" tag is the status code when below 200 or above 399 https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java

packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
russell-matt: Sure, I'll make it immutable and inside the scoped block. Do you want it to handle codes <200 like how Brave does, or keep it to being 400 and greater only?

eirslett:
Is it possible to add a unit test as well?

adriancole:
thanks a lot!

PR #222, opened 2018-06-28
[#197] Adds support for sampling flags extraction.

### LINKED ISSUE #197: Support SamplingFlags extraction?
`brave` and `zipkin-php` include the concept of `SamplingFlags` which allows propagation to extract sampling decisions from headers although `trace` information is not there. Shall we support something like that here?

At the moment when such a thing happens, `hapi middleware` fails to extract `traceId`.

Ping @adriancole

jcchavezs:
This PR aims to add support for tracing flags in zipkin-js.

Ping @adriancole @fbaiodias @pchiwan

Closes #197

### DIFF
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -59,7 +59,7 @@ class HttpServerInstrumentation {
           parentId: None,
           spanId: new Some(currentId.spanId),
           sampled: readHeader(Header.Sampled) === None ?
-              currentId.sampled : readHeader(Header.Sampled),
+              currentId.sampled : readHeader(Header.Sampled).map(stringToBoolean),
           flags: readHeader(Header.Flags) === None ? currentId.flags : readHeader(Header.Flags),
         });
         return new Some(idWithFlags);
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -52,14 +52,15 @@ class HttpServerInstrumentation {
         });
       });
     } else {
-      if (readHeader(Header.Flags) !== None) {
+      if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
         const currentId = this.tracer.id;
         const idWithFlags = new TraceId({
-          traceId: currentId.traceId,
-          parentId: currentId.parentId,
-          spanId: currentId.spanId,
-          sampled: currentId.sampled,
-          flags: readHeader(Header.Flags)
+          traceId: new Some(currentId.traceId),
+          parentId: None,
+          spanId: new Some(currentId.spanId),
+          sampled: readHeader(Header.Sampled) === None ?
+              currentId.sampled : readHeader(Header.Sampled),
+          flags: readHeader(Header.Flags) === None ? currentId.flags : readHeader(Header.Flags),
         });
         return new Some(idWithFlags);
       } else {
--- a/packages/zipkin/test/httpServerInstrumentation.test.js
+++ b/packages/zipkin/test/httpServerInstrumentation.test.js
@@ -121,6 +121,8 @@ describe('Http Server Instrumentation', () => {
         {'X-B3-Flags': '1'},
         {'X-B3-Sampled': '0'},
         {'X-B3-Sampled': '1'},
+        {'X-B3-Sampled': 'true'},
+        {'X-B3-Sampled': 'false'},
         {'X-B3-Sampled': '0', 'X-B3-Flags': '0'},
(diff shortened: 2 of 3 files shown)


packages/zipkin/test/httpServerInstrumentation.test.js
```diff
@@ -107,6 +107,55 @@ describe('Http Server Instrumentation', () => {
...
+      expect(annotations[2].annotation.annotationType).to.equal('BinaryAnnotation');
+      expect(annotations[2].annotation.key).to.equal('http.url');
+      expect(annotations[2].annotation.value).to.equal(url);
+
+      expect(annotations[3].annotation.annotationType).to.equal('ServerRecv');
+
+      expect(annotations[4].annotation.annotationType).to.equal('LocalAddr');
+
+      expect(annotations[5].annotation.annotationType).to.equal('BinaryAnnotation');
+      expect(annotations[5].annotation.key).to.equal('X-B3-Flags');
```
adriancole: just noticing.. this is weird... do we literally capture headers as tags (aka binary annotations?)

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -51,14 +51,14 @@ class HttpServerInstrumentation {
...
+      if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
         const currentId = this.tracer.id;
         const idWithFlags = new TraceId({
-          traceId: currentId.traceId,
-          parentId: currentId.parentId,
-          spanId: currentId.spanId,
-          sampled: currentId.sampled,
-          flags: readHeader(Header.Flags)
+          traceId: new Some(currentId.traceId),
+          parentId: new Some(currentId.parentId),
```
adriancole: is some right here? Can't parentId be absent?

adriancole: looks like the current behavior returns the span ID when the parent is absent, which is kindof surprising.

your code will work, but I think it is walking around a general flaw here. parentId return an Option even if it is consistently not an option today (ex in typescript https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/index.d.ts#L61)

@eirslett need a second opinion on this which amounts to a separate issue I suppose if the behavior is not desirable. Note: I've had to do some defense in the UI as sometimes people have written json with spanId == parentId accidentally.

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -51,14 +51,14 @@ class HttpServerInstrumentation {
         });
       });
     } else {
-      if (readHeader(Header.Flags) !== None) {
+      if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
```
adriancole: I think this change needs a test case... looks like the intent is to also allow 'X-B3-Sampled' without trace identifiers... correct? similar to https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java#L125

packages/zipkin/test/httpServerInstrumentation.test.js
jcchavezs: It is, I could remove it but since this is a copy fromt he other test I would prefer to keep it an remove in a next PR.

adriancole:
thanks yeah doing separate things in PRs is awesome. I just noticed a
couple things that looked surprising but didnt imply addressing them in
this commit.

packages/zipkin/src/instrumentation/httpServer.js
jcchavezs: I added tests for both: https://github.com/openzipkin/zipkin-js/pull/222/files#diff-69d935ef65c95d1891e05297e13f46aaR118

packages/zipkin/test/httpServerInstrumentation.test.js
```diff
@@ -105,6 +105,67 @@ describe('Http Server Instrumentation', () => {
...
+
+    const port = 80;
+    const url = `http://127.0.0.1:${port}`;
+    const instrumentation = new HttpServer({tracer, serviceName: 'service-a', port});
+
+    const headersCases = [
+        {},
+        {'X-B3-Flags': '0'},
+        {'X-B3-Flags': '1'},
+        {'X-B3-Sampled': '0'},
```
adriancole: hmm this shouldn't result in recording actually... right?

jcchavezs:
@adriancole yeah, as I said, something should definitively check it before reporting but I was not sure whether this had same behaviour as brave and zipkin-php that when it is not sampled, we simply don't record any annotation.

adriancole:
cool.. merging.

PR #224, opened 2018-06-28
[#57] Removes http header binary annotation in spans.

jcchavezs:
This PR removes the `'X-B3-Flags'` from the binary annotations (aka tags): actual removal is in https://github.com/openzipkin/zipkin-js/compare/removes_unnecessary_debug_tag?expand=1#diff-13acaf8526b6f17f74c05c209a6c833c and the rest are tests.

Ping @adriancole @eirslett

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 lib
 npm-debug.log
 lerna-debug.log
+package-lock.json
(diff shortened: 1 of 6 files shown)


adriancole:
cool on non appveyor green

Issue #217, opened 2018-06-01
Remove http.url in favor of http.path in express middleware


jcchavezs:
At the moment, Zipkin JS is including `http.url` as annotation including all query strings. This has som privacy issues (e.g. passing e-mails or tokens in query parameters) so most zipkin instrumentations include the `http.path` instead.

- [zipkin-go](https://github.com/openzipkin/zipkin-go/blob/e260ab9d9af33ad4d9267be78add1d8a62806f74/middleware/http/server.go#L95)
- [zipkin-php](https://github.com/jcchavezs/zipkin-instrumentation-symfony/blob/master/src/ZipkinBundle/Middleware.php#L81)
- [zipkin-ruby](https://github.com/openzipkin/zipkin-ruby/blob/e2f429976a7bfddbe908743bc2c9c2ec0b64afb4/CHANGELOG.md#0184)

Ping @pchiwan @adriancole

adriancole:
sounds ok just need to be clear how folks who currently rely on this can
transition or otherwise affect tags

PR #226, opened 2018-06-29
[Issue #217] Remove http.url annotation in favor of http.path and http.host

pchiwan:
At the moment, Zipkin JS is including `http.url` as annotation including all query strings. This has some privacy issues (e.g. passing e-mails or tokens in query parameters) so most zipkin instrumentations include the `http.path` instead.

Ping @jcchavezs @adriancole


adriancole:
if privacy is the main concern totally agree. however adding http.host isnt
usual.

why don't we start with http.path? that way we dont add 2 potentially high
cardinality tags in place of one

jcchavezs:
:+1: Amazing

jcchavezs:
Ping @eirslett

adriancole:
Thanks so much!

Issue #210, opened 2018-05-12
Express Middleware not reporting errors on 5xx


chris-smith-zocdoc:
The express middleware is not currently marking 4xx/5xx server response with the error annotation. This is behavior is present in other zipkin libs [like brave](https://github.com/openzipkin/brave/blob/b3c33ff0146d4de96bea251cb3659358155566d8/instrumentation/http/README.md)

I believe the issue is that the [httpServer](https://github.com/openzipkin/zipkin-js/blob/6a12a1d3bb0567e7e3e3e69a69a1c4bffdeb0351/packages/zipkin/src/instrumentation/httpServer.js#L86) takes `error` as a third argument but the [express middleware](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-express/src/expressMiddleware.js#L35) only calls it with the spanId and status code. 

One potential solution would be to remove the third argument from httpServer.recordResponse and just infer if there was an error from the status code.

jcchavezs:
@chris-smith-zocdoc we usually check the status code to determine whether there was an error. Any chance you can open a PR for this?

chris-smith-zocdoc:
@jcchavezs Sure, I just put one up

PR #233, opened 2018-07-06
record an error if there was one in the express middleware

chris-smith-zocdoc:
Fixes https://github.com/openzipkin/zipkin-js/issues/210

### DIFF
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -32,7 +32,9 @@ module.exports = function expressMiddleware({tracer, serviceName, port = 0}) {
 
       res.on('finish', () => {
         tracer.scoped(() => {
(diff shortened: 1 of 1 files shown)


chris-smith-zocdoc:
Added a test

chris-smith-zocdoc:
If we have more information I agree we could add it, I was just following what the other libs are already doing.

[Brave](https://github.com/openzipkin/brave/blob/b3c33ff0146d4de96bea251cb3659358155566d8/instrumentation/http/src/test/java/brave/http/HttpParserTest.java#L51-L52) sometimes sets the status code, sometimes the exception. 

[Restify](https://github.com/openzipkin/zipkin-js/blob/4e8fdcccc5a5fba08044d070fb99ef6de9de93be/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js#L39-L44) sets the status code

jcchavezs:
Thanks @chris-smith-zocdoc

PR #234, opened 2018-07-07
change addZipkinHeaders method to not modify the IncomingMessage object

Eli-Goldberg:
An incoming request is actually an [IncomingMessage](https://nodejs.org/api/http.html#http_class_http_incomingmessage) object.
This fixes any attempt to use the on() event method after zipkin modifies it's headers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ npm-debug.log
 lerna-debug.log
 yarn-error.log
 package-lock.json
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ npm-debug.log
 lerna-debug.log
 yarn-error.log
 package-lock.json
-.vscode/
--- a/packages/zipkin/src/request.js
+++ b/packages/zipkin/src/request.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-param-reassign */
 const HttpHeaders = require('./httpHeaders');
 
 function appendZipkinHeaders(req, traceId) {
@@ -16,8 +17,9 @@ function appendZipkinHeaders(req, traceId) {
 }
 
 function addZipkinHeaders(req, traceId) {
-  const headers = appendZipkinHeaders(req, traceId);
-  return Object.assign({}, req, {headers});
+  req.headers = appendZipkinHeaders(req, traceId);
+
+  return req;
 }
 
 module.exports = {
--- a/packages/zipkin/test/request.test.js
+++ b/packages/zipkin/test/request.test.js
@@ -2,8 +2,22 @@ const Request = require('../src/request.js');
 const HttpHeaders = require('../src/httpHeaders');
 const {Some} = require('../src/option');
 const TraceId = require('../src/tracer/TraceId');
+const {IncomingMessage} = require('http');
 
 describe('Request', () => {
+  it('should not change an object\'s existing methods', () => {
+    const traceId = new TraceId({
+      traceId: new Some('48485a3953bb6124'),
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ npm-debug.log
 lerna-debug.log
 yarn-error.log
 package-lock.json
+.vscode/
```
jcchavezs: Since this is IDE specific and not something related to the library I'd rather include this in the global `.gitignore`. See https://gist.github.com/subfuzion/db7f57fff2fb6998a16c.

packages/zipkin/src/request.js
```diff
@@ -16,8 +17,9 @@ function appendZipkinHeaders(req, traceId) {
 }
 
 function addZipkinHeaders(req, traceId) {
-  const headers = appendZipkinHeaders(req, traceId);
-  return Object.assign({}, req, {headers});
+  req.headers = appendZipkinHeaders(req, traceId);
+
+  return req;
```
jcchavezs: Request is supposed to be immutable, no middleware is supposed to change a request. Could you please explain the intention behind this change?

packages/zipkin/test/request.test.js
```diff
@@ -2,8 +2,22 @@ const Request = require('../src/request.js');
...
+  it('should not change an object\'s existing methods', () => {
+    const traceId = new TraceId({
+      traceId: new Some('48485a3953bb6124'),
+      spanId: '48485a3953bb6124'
+    });
+
+    const testRequest = new IncomingMessage();
+    const resultRequest = Request.addZipkinHeaders(testRequest, traceId);
+
+    expect(resultRequest.on).to.exist; // eslint-disable-line no-unused-expressions
```
jcchavezs: I've seen this comment in a previous file. Is it really needed?

Eli-Goldberg:
Sorry, I should have started with a background explanation.
I tried to make it short :)

I came accross this when tried to use the [appmetrics-zipkin](https://github.com/RuntimeTools/appmetrics-zipkin) in [Koa](https://github.com/koajs/koa).

_appmetrics-zipkin_ creates hooks for every incoming and outgoing http requests,
and replaces the request with one that has zipkin headers and sends that information to a zipkin server.

In general everything works, but the problems is that when using Koa, you pretty much have to use koa-bodyparser for anything other than get requests, and it uses the events on every incoming request (which is a node IncomingMessage object).

Since appmetrics [replaces](https://github.com/RuntimeTools/appmetrics-zipkin/blob/master/probes/http-probe-zipkin.js#L116) the entire object with the result of addZipkinHeaders, the IncomingMessage is replaced with a shallow copy (Object.assign) of it, losing the on() methods.

I completely agree with the idea of immutability,
but since you are replacing only the headers part - why not just have the function return a copy of the merged headers?
Or on the other hand simply have the function mutate only the headers of the request (would be much faster than creating a copy) - [https://runkit.com/eli-goldberg/object-assign-speed](https://runkit.com/eli-goldberg/object-assign-speed)

In any case it sounds like it's appmetric's fault and they shouldn't modify the entire request, but rather only take the headers from addZipkinHeaders and use that.

Eli-Goldberg:
@jcchavezs I've opened a PR to appmetrics: https://github.com/RuntimeTools/appmetrics-zipkin/pull/32
So feel free to close this PR if you feel as if this one isn't needed :)
Thanks

jcchavezs:
@Eli-Goldberg thanks for taking the time to do the rationale behind this. We really appreciate it.

Since you took the time to open a PR for this I hope this won't be the last time.

Issue #236, opened 2018-07-10
Allow httpServer instrumentation host to be overridden


rcoedo:
Currently the httpServer instrumentation implementation is querying the OS to find the host's ip address on every request. This can lead to a performance problem when the server is under heavy load.

![performance](https://user-images.githubusercontent.com/6409659/42496292-d341ba60-8425-11e8-83b3-627601956214.png)

Allowing the host to be overridden is a simple fix for this problem and it wouldn't break the API.

jcchavezs:
Closed by https://github.com/openzipkin/zipkin-js/pull/237

PR #237, opened 2018-07-10
Allow host to be overridden in httpServer instrumentation (#236)

rcoedo:
Link to the issue: https://github.com/openzipkin/zipkin-js/issues/236

### DIFF
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -1,8 +1,9 @@
-const Header = require('../httpHeaders');
-const {Some, None} = require('../option');
-const TraceId = require('../tracer/TraceId');
 const Annotation = require('../annotation');
+const Header = require('../httpHeaders');
+const InetAddress = require('../InetAddress');
+const TraceId = require('../tracer/TraceId');
 const parseRequestUrl = require('../parseUrl');
+const {Some, None} = require('../option');
 
 function stringToBoolean(str) {
   return str === '1' || str === 'true';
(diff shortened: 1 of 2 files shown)


jcchavezs:
Thanks for doing this @rcoedo. Would you mind to add tests for this?

rcoedo:
Not at all, I will update the PR as soon as possible.

jcchavezs:
Awesome work @rcoedo I am only concerned on not being breaking the default behavior. Can we cover it with a test as well? 

After that I believe we are ready to merge. Any feedback @adriancole?

rcoedo:
@jcchavezs I'm afraid the test would be dependent on the host ip if I try to check the `LocalAddr` annotation ip address, and if I don't then the test will only check that the instrumentation records a `LocalAddr` annotation, just like all the other tests that we have.

We can't mock the "network-address" library with our current tooling, can we? Any other ideas? 😄

rcoedo:
Maybe it is a bit ugly to query the OS as a side effect of instantiating a `LocalAddr` annotation, but I'm not sure about the implications of moving this behaviour out of it. I guess it would be a breaking change if users are instantiating LocalAddr annotations explicitly.

jcchavezs:
I think it is good enough for now if we test that the annotation is there. This will guarantee that the flow is not breaking if we don't pass the parameter. Does it make sense to you?

rcoedo:
This was is already being done by literally every other test but it is fine with me 👍 

If you have any other thoughts or suggestions just let me know! 😄

jcchavezs:
Great work!

PR #239, opened 2018-07-10
[#197] Improves sampling flags testing with meaningful cases.

jcchavezs:
Ping @pchiwan @adriancole

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -151,15 +151,20 @@ class Tracer {
   }
 
   recordAnnotation(annotation) {
+    let shouldRecord = this.id.flags === 1;
+
     this.id.sampled.ifPresent(sampled => {
       // sampled present is different than sampled == true
-      if (!sampled) return;
+      shouldRecord = sampled;
+    });
+
+    if (shouldRecord) {
       this.recorder.record(new Record({
         traceId: this.id,
         timestamp: now(this._startTimestamp, this._startTick),
         annotation
       }));
-    });
+    }
   }
 
   recordMessage(message) {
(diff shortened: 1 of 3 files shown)


jcchavezs:
Ping @DanielMSchmidt @yonran

packages/zipkin/src/tracer/index.js
```diff
@@ -151,15 +151,20 @@ class Tracer {
   }
 
   recordAnnotation(annotation) {
+    let shouldRecord = this.id.flags === 1;
```
jcchavezs: @adriancole I am not entirely sure why we were not checking the debug flags for recording spans. Any input?

packages/zipkin/test/httpServerInstrumentation.test.js
```diff
@@ -166,6 +118,89 @@ describe('Http Server Instrumentation', () => {
     });
   });
 
+  const samplingFlagCases = [
```
jcchavezs: These are all possible header combinations we should support.

packages/zipkin/test/httpServerInstrumentation.test.js
```diff
@@ -166,6 +118,89 @@ describe('Http Server Instrumentation', () => {
...
+        {headers: {}, hasAnnotations: null},
+        {headers: {'X-B3-Flags': '0'}, hasAnnotations: null},
```
jcchavezs: `hasAnnotations: null` means that we can not determine whether the annotation is recorded or not (depends on sampling policy). Although we could inject an `alwaysSampler` it is not the question of this test.

packages/zipkin/src/tracer/index.js
adriancole: debug needs to end up as span.debug when reporting. In order for the trace to be consistently debug (ex when propagating downstream and at collector tier), it needs to be carried somehow (flag or otherwise in the context), and also written as span.debug=true in json (note we don't ever write span.debug=false)

PR #245, opened 2018-07-12
Relax response code expectation to allow 200s

keitwb:
We have mimicked the Zipkin API but don't use 202s since they are not appropriate in our case.

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -60,7 +60,7 @@ class HttpLogger extends EventEmitter {
         headers: self.headers,
         timeout: self.timeout,
       }).then((response) => {
-        if (response.status !== 202) {
+        if (response.status !== 202 && response.status !== 200) {
           const err = 'Unexpected response while sending Zipkin data, status:' +
             `${response.status}, body: ${postBody}`;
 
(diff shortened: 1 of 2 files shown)


adriancole:
not sure we have a unit test, but if there is, mind making sure this stays fixed? While yeah not per the [swagger](https://zipkin.io/zipkin-api/#/default/post_spans) sounds reasonable.

packages/zipkin-transport-http/src/HttpLogger.js
```diff
@@ -60,12 +60,14 @@ class HttpLogger extends EventEmitter {
...
       }).then((response) => {
-        if (response.status !== 202) {
+        if (response.status !== 202 && response.status !== 200) {
           const err = 'Unexpected response while sending Zipkin data, status:' +
             `${response.status}, body: ${postBody}`;
 
           if (self.errorListenerSet) this.emit('error', new Error(err));
           else console.error(err);
+        } else {
+          this.emit('success', response);
```
keitwb: Couldn't find any way to test successful response handling on the logger's part without having some kind of positive feedback for the test to hook into.

keitwb: Adding an error handler and then calling `expect.fail()` in that was just making it raise unhandled promise errors and wouldn't actually fail.

adriancole:
This is not a universally popular idea, but I think it is reasonable to be lenient. This transport is not a compatibility check, and being very strict is imho worse for users who may have their own auth proxies they also can't control.

If folks feel like reverting this, I wouldn't argue terribly, but I think it was a reasonable call to merge it.

PR #246, opened 2018-07-13
fix: kipkin ConsoleRecorder constructor does not match ts type

tomasAlabes:
I adapted the Typescript definition file to match the implementation. I didn't change the implementation because I don't know what it's the intention of the developers. `BatchRecorder` receives a `Logger` type.
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/console-recorder.js#L8
vs
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L59

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -269,7 +269,7 @@ declare namespace zipkin {
   }
 
   class ConsoleRecorder implements Recorder {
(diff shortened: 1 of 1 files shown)


jcchavezs:
Hi @tomasAlabes thanks for doing this! I was wondering if we could somehow add a check to prevent that further changes don't match the implementation.

PR #254, opened 2018-08-10
Fix TS compilation error for missing Option type argument

kevin-greene-ck:


jcchavezs:
@kevin-greene-ck thanks for this! This PR is related to https://github.com/openzipkin/zipkin-js/issues/252. Would you be up for adding the compilation as a travis step so we avoid breaking the TS scripts with further changes? It can be in a next PR if you prefer.

kevin-greene-ck:
@jcchavezs Yeah, no problem. I can probably get to that over the weekend.

jcchavezs:
Awesome @kevin-greene-ck!

Issue #263, opened 2018-09-21
addZipkinHeaders return void instead of object


sonybinhle:
In the index.d.ts file, the addZipkinHeaders return void
```javascript
namespace Request {
    function addZipkinHeaders(req: {headers: any}, traceId: TraceId): void;
}
```

The actual implementation is the function addZipkinHeaders return a object
```javascript
function addZipkinHeaders(req, traceId) {
  const headers = appendZipkinHeaders(req, traceId);
  return Object.assign({}, req, {headers});
}
```

Same as https://github.com/openzipkin/zipkin-js/issues/190

jcchavezs:
`addZipkinHeaders(req, traceId)` is supposed to return a new request so could you please open a PR changing it?

On the other hand, there is [this issue](https://github.com/openzipkin/zipkin-js/issues/248) that if resolved would avoid such differences.

sonybinhle:
@jcchavezs I would like to create a PR for this issue.
If there is no one work on #248 I would also like to work on it

jcchavezs:
Awesome!

Den fre. 21. sep. 2018, 15:10 skrev Binh Le <notifications@github.com>:

PR #264, opened 2018-09-21
#263: Change return type of addZipkinHeaders to RequestZipkinHeaders

sonybinhle:
I tried to find the coding convention for implementing the fix but it may be not available. Please give me feedback so I could modify my PR.


jcchavezs:
Hi @sonybinhle thanks a lot for this! I am curious about this change and how we could test this before we merge it.

Ping @kevin-greene-ck @0v3rst33r

jcchavezs:
@sonybinhle can something like this work? https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files

sonybinhle:
@jcchavezs I rebased my PR to master.
About the "type-checking-javascript-files", I will research it tonight.
And this is how lodash write test to check definition types for their code:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c0912e8b266f737d94e46493b76066df377170de/types/lodash/v3/lodash-tests.ts

It just merely write some test on typescript, runtime typescript will verify the return object to assure they match with definition types.

jcchavezs:
Awesome @sonybinhle looking forward to see the approach.

sonybinhle:
@jcchavezs I can summary my research about the approaches:

1. Type Checking JavaScript Files:
It is for checking *.js file using typescript by introducing commented metadata types. The purpose of this is adding more type-safety for the *.js files and doesn't support verifying typescript definitions APIs, in our case index.d.ts file.

2. The reason we wrote index.d.ts is for exporting/documenting our library public APIs used by other typescript files. Lodash definition files are tested by **dtslint** which verify the definition files. The verification is very shallow, it is like a double check. For example, if in index.d.ts we said function A return string, **dtslint** verify it again as linting(**compiling time**) like `const x: string = A()`. IMO, it adds very little value since it couldn't verify against the real return value of the real function used, the reason is that typescript remove types at **runtime**.

3. index.d.ts is looked like documentation. Let imagine we write documentation for our function, we say a function A return a string, if later we change the function's signature to return a number, we could forget to update our documentation, later the issue could be found out and we updated the documentation. I mean the process of synchronizing between code and documentation is mostly done manually(by the human). So should we do it manually(not writing tests at all) in our case the file index.d.ts?

I think that in order to make sure our index.d.ts is compatible with our real lib's APIs, we could write test using ts, keep their types during runtime, and compare it against types definition in index.d.ts.
I didn't see any approach like that before, so I am worried it may be a wrong way. Could you please give me some advice?

jcchavezs:
> I think that in order to make sure our index.d.ts is compatible with our real lib's APIs, we could write test using ts, keep their types during runtime, and compare it against types definition in index.d.ts.
I didn't see any approach like that before, so I am worried it may be a wrong way. Could you please give me some advice?

I think it makes totally sense @sonybinhle. I am not an expert on TS (nor JS) but I think what you propose could work, also I saw that [TS can also compile JS](https://github.com/Microsoft/TypeScript/issues/14536) files so what I think we could also do to avoid writing silly tests is:

```js
// file.js
do_something_impl(a){
    if (...) {
        return false
    }
    return "abc"
}
```

```typescript
//file.ts
function do_something(string a): string
```
then we on-the-flight create the wrapper:

```typescript
function do_something(string a): string {
    return do_something_impl(a)
}
```

and run the linter which will complain if the do_something_impl(a) does not match the actual signature.

I know this is more complex than writing the actual TS tests so I guess you are the best one to make the call on how to write the first test.

We can start with one test for this specific case and then see if we can extend it to all of them in the future. Does it makes sense?

sonybinhle:
@jcchavezs 
I updated the PR and proposed a types test strategy followed what we discussed.
The summary of what I did:

1. Using **ts-runtime**, the typescript compiler which retains types after compiling ts to js.
 **ts-runtime** is the wrapper of **flow-runtime** which developed with the same purpose for runtime-type-checking.
2. Write test in **test-types/*.test.ts** using mocha, chai syntax.
3. **ts-runtime** compile ts to js and also add additional javascript for types checking.
4. Mocha run test files as normal with profile mocha-types.opts

I catch some type bugs on my previous implementation by using the above strategy, we can check ourself by removing the character 'X' from 'X-B3-TraceId'.

jcchavezs:
This is atonishing work @sonybinhle! Thanks a lot.

You might want to write a blog post for this somewhere.

Den ons. 3. okt. 2018, 17:05 skrev Binh Le <notifications@github.com>:

jcchavezs:
@sonybinhle travis for node <4 is returning:

```
error ts-runtime@0.1.35: The engine "node" is incompatible with this module. Expected version ">=6".
```

Maybe we need to use a former version? Otherwise maybe we can just launch these tests for node > 6.

sonybinhle:
@jcchavezs IMO, the type-tests or type checking are only for checking the type definitions which used by typescript project. So it definitely **will not affect** our lib production bundle or it **does not have any influence** on how well our bundle runs on Node 4. So I thought that we could run types test **only on Node 8**. 
But I didn't find any good solution for installing **ts-runtime** only when using **Node 8**.

One of the ugly solutions is using  `yarn install --ignore-engines`.
Another ugly solution is instead of using `yarn install` directly. We could write `installScript.js` and modify `package.json` based on the Node version before running `yarn install`.

I saw that the above solutions add more complexity. How do you think about it?

jcchavezs:
Makes totally sense. I did not know ts-runtime only supported node >6. I am
trying now to ignore platform request for node 4. Another option is to just
uninstall ts-runtime before installing all dependencies only for node 4.

Den lør. 6. okt. 2018, 08:13 skrev Binh Le <notifications@github.com>:

sonybinhle:
@jcchavezs I saw your commit, thank you very much for updating travis.yml

packages/zipkin/index.d.ts
```diff
@@ -287,8 +287,17 @@ declare namespace zipkin {
     letContext<V>(ctx: TraceId, callback: () => V): V;
   }
 
+  type RequestZipkinHeaders<T = any, H = any> = T & {
+    headers: H & {
+      ['X-B3-TraceId']: string;
+      ['X-B3-SpanId']: string;
+      ['X-B3-ParentSpanId']?: string;
+      ['X-B3-Sampled']?: '1' | '0';
```
adriancole: ps and assuming this is server-side:
* we often leniently parse 'true' and 'false' for old clients, as they used to send this.
* we'll want to update this for 'b3' header once that chore is done.

adriancole: also, I think this is missing `['X-B3-Flags']?: '1' | '0';` eventhough '0' is redundant and should not be sent.. some might send it. Flag 1 is "debug"

jcchavezs:
@sonybinhle could you please address @adriancole's comments in a new PR?

packages/zipkin/index.d.ts
sonybinhle: Hi @adriancole, nice to meet you, thank you very much for your suggestions. I paste the content of the current **request.js** file below:
```
const HttpHeaders = require('./httpHeaders');

function appendZipkinHeaders(req, traceId) {
  const headers = req.headers || {};
  headers[HttpHeaders.TraceId] = traceId.traceId;
  headers[HttpHeaders.SpanId] = traceId.spanId;

  traceId._parentId.ifPresent(psid => {
    headers[HttpHeaders.ParentSpanId] = psid;
  });
  traceId.sampled.ifPresent(sampled => {
    headers[HttpHeaders.Sampled] = sampled ? '1' : '0';
  });

  return headers;
}

function addZipkinHeaders(req, traceId) {
  const headers = appendZipkinHeaders(req, traceId);
  return Object.assign({}, req, {headers});
}

module.exports = {
  addZipkinHeaders
};
```
When I read the code above, I saw only HttpHeaders.TraceId, HttpHeaders.SpanId, HttpHeaders.ParentSpanId, HttpHeaders.Sampled and headers[HttpHeaders.Sampled] is either '1' or '0'.

> we often leniently parse 'true' and 'false' for old clients, as they used to send this.
It might not affect old clients, because the index.d.ts file is only for typescript projects to use type inference at compile time, so it does not influence old-clients runtime.

>  I think this is missing ['X-B3-Flags']?: '1' | '0';
I couldn't find ['X-B3-Flags'] on the source code above, but if it is **optional(with ?)** and not presented on the source code, we might ignore it.

My response is only based on my limited knowledge about the project; please clarify me if I miss-understood something.

adriancole:
The problem is that if someone is allowed to send X-B3-Flags: 1 and
omit X-B3-Sampled in this case (as Flags: 1 implies sampled).

So I suppose I noticed an old bug..
https://github.com/openzipkin/b3-propagation#sampling-state

jcchavezs:
Yeah, `X-B3-Flags` is supposed to be there, being optional and I believe if
one sends `X-B3-Flags` there is no need to send `X-B3-Sampled` right
@adriancole?

José Carlos Chávez


Den tir. 9. okt. 2018 kl. 02:08 skrev Adrian Cole <notifications@github.com
>:

> The problem is that if someone is allowed to send X-B3-Flags: 1 and
> omit X-B3-Sampled in this case (as Flags: 1 implies sampled).
>
> So I suppose I noticed an old bug..
> https://github.com/openzipkin/b3-propagation#sampling-state

adriancole:
>
>  I believe if one sends `X-B3-Flags` there is no need to send
> `X-B3-Sampled` right
>
yep. it is fine to send both, but omitting X-B3-Sampled is fine.
X-B3-Flags: 0 is ignored in any case

>

sonybinhle:
@adriancole @jcchavezs Sorry but after reading your conversation many times, I still didn't get the idea. :grin:

Could you please give me some hints of how to update it properly?

jcchavezs:
Sure @sonybinhle. In sum: `X-B3-Flags` is missing, in the headers and should be added as optional.

Reagarding 
```js
function appendZipkinHeaders(req, traceId) {
  const headers = req.headers || {};
  headers[HttpHeaders.TraceId] = traceId.traceId;
  headers[HttpHeaders.SpanId] = traceId.spanId;

  traceId._parentId.ifPresent(psid => {
    headers[HttpHeaders.ParentSpanId] = psid;
  });
  traceId.sampled.ifPresent(sampled => {
    headers[HttpHeaders.Sampled] = sampled ? '1' : '0';
  });

  return headers;
}
```

We are lacking:

```js
  traceId.flags.ifPresent(flags => {
    if (flags == '1') {headers[HttpHeaders.Flags]='1'}
  });
```

and indeed, if `flags` is present and it is `'1'`, we should not send the sample. i.e. avoid the 
```js
  traceId.sampled.ifPresent(sampled => {
    headers[HttpHeaders.Sampled] = sampled ? '1' : '0';
  });
```

-----

**In the future**, we could check that when `traceId` is populated, we only add `flags` as `'1'` and then 

```js
  traceId.flags.ifPresent(flags => {
    if (flags == '1') {headers[HttpHeaders.Flags]='1'}
  });
```

becomes 

```js
  traceId.flags.ifPresent(flags => {
    headers[HttpHeaders.Flags] = '1'
  });
```
but lets go step by step. cc @adriancole

jcchavezs:
@sonybinhle did you have time to have a look at this? Thanks!

sonybinhle:
@jcchavezs Thank you very much for reminding me. I will start working on #268 and fixing it there.

PR #269, opened 2018-10-19
Removing require.call

### LINKED ISSUE #195: Zipkin warnings with create-react-app
I created a react app with **create-react-app** and when i import zipkin i get this error on start and build.

```
./node_modules/zipkin-transport-http/lib/HttpLogger.js
17:27-34 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/zipkin/lib/InetAddress.js
62:23-30 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

I put this code in App.js after all the imports.
The trace is working an all the requests appear in zipkin server, but i need to clear this warnings so it passes on the pipeline.

```javascript
const {
  Tracer,
  BatchRecorder,
  jsonEncoder: {JSON_V2}
} = require('zipkin');
const CLSContext = require('zipkin-context-cls');
const {HttpLogger} = require('zipkin-transport-http');

// Setup the tracer to use http and implicit trace context
const tracer = new Tracer({
  ctxImpl: new CLSContext('zipkin'),
  recorder: new BatchRecorder({
    logger: new HttpLogger({
      endpoint: 'http://localhost:9411/api/v2/spans',
      jsonEncoder: JSON_V2
    })
  }),
  localServiceName: 'service-a' // name of this application
});

// now use tracer to construct instrumentation! For example, fetch
//https://randomuser.me/api
const remoteServiceName = 'randomuser.me';
const zipkinFetch = wrapFetch(fetch, {tracer, remoteServiceName});
```

And my package is 

```javascript
    "zipkin": "^0.12.0",
    "zipkin-context-cls": "^0.11.0",
    "zipkin-instrumentation-fetch": "^0.12.0",
    "zipkin-transport-http": "^0.12.0"
```

DavideCarvalho:
Since require.call it's making webpack give some warnings, I'm removing it and using only require. The change was made on "INetAddress" on zipkin package and on "HttpLogger" on zipkin-transport-http package.
The tests continue passing as it should.
Closes #195

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -4,7 +4,7 @@ const globalFetch =
   (typeof global !== 'undefined' && global.fetch);
 
 // eslint-disable-next-line global-require
-const fetch = globalFetch || require.call(null, 'node-fetch');
+const fetch = globalFetch || require('node-fetch');
 
 const {
   jsonEncoder: {JSON_V1}
--- a/packages/zipkin/package.json
+++ b/packages/zipkin/package.json
@@ -14,8 +14,7 @@
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
     "base64-js": "^1.1.2",
-    "is-promise": "^2.1.0",
-    "network-address": "^1.1.0"
+    "is-promise": "^2.1.0"
   },
   "devDependencies": {
     "babel-cli": "^6.23.0",
--- a/packages/zipkin/package.json
+++ b/packages/zipkin/package.json
@@ -14,7 +14,8 @@
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
     "base64-js": "^1.1.2",
-    "is-promise": "^2.1.0"
+    "is-promise": "^2.1.0",
+    "network-address": "^1.1.0"
   },
   "devDependencies": {
     "babel-cli": "^6.23.0",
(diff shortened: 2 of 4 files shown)


jcchavezs:
Thanks a lot @DavideCarvalho!

Do you have any comments on this @DanielMSchmidt @Labrat02? otherwise I will just merge it.

Labrat02:
@jcchavezs Thanks a lot Jose.  Just so you know, I pulled down the codebase and tested this exact change months ago and it worked perfectly for us.  So, no I have zero problems with the change.  We ended up making changes to suppress the hide these specific warnings so that our builds would not fail so honestly this issue no longer blocks us, however it will hopefully help other people who use React and webpack in CI.

FYI, since require.call was only used to solve an issue on the NodeJS side, that is what you really need to be mindful of.  We don't use the library in any Node applications so I don't know if there is some issue that might be introduced on that side.

Thanks again!  Merge Away!

DanielMSchmidt:
I don't recall this change exactly, but if you tested it out in the browser and you are sure this doesn't break compatability with using it with something that is not webpack like browserify, please go ahead and merge

DavideCarvalho:
I’ve tested only with webpack, but if you guys need testing with other bundlers, I can do it

adriancole:
maybe test browserify and call it done?

DavideCarvalho:
I was testing with Webpack and sadly, I'm getting the error on InetAddress.js even without ```require.call```
Since zipkin-js use only one method from network-address, and this lib is not updated for 2 years, I'm thinking about getting the lib code and put it on zipkin-js.
What you guys think?

jcchavezs:
Great work @DavideCarvalho, do you think it is possible to add a unit test so we make sure we don't break browserify nor webpack with a further change?

DavideCarvalho:
Okay, so i've made some manual testing with webpack. The warnings are off.
I'll test browserify once i get home.

jcchavezs:
Great job @DavideCarvalho!

Issue #268, opened 2018-10-08
Backfill ts tests for actual implementations.


jcchavezs:
After https://github.com/openzipkin/zipkin-js/pull/264 was merged now we have a way to test implementation to fulfill TS signature. What is missing now is to backfill tests for all packages.

Ping @sonybinhle

sonybinhle:
@jcchavezs Could you please assign this issue to me?

What is the standard PR styles in our project?
For example, I prefer to write the test for about 10 functions at once and file by file.
I mean if I create the PR for this issue, I will divide it into multiple small PRs.

adriancole:
usually we do a PR per task not per file. if all the changes are related I
would raise one pr

sonybinhle:
@adriancole @jcchavezs Thank you very much, so I will raise the PRs per package.

PR #271, opened 2018-10-21
Backfill ts tests for actual implementations. #268

sonybinhle:
I prepared the type test files based on our functional test files, please give me feedback.


packages/zipkin/index.d.ts
```diff
@@ -239,7 +239,7 @@ declare namespace zipkin {
       constructor(key: string, value: boolean | string | number);
       readonly annotationType: string;
       key: string;
-      value: string;
+      value: boolean | string | number;
```
jcchavezs: I don't think we should change this, annotations should be only strings as that is what zipkin supports.

jcchavezs: I think the last one is enough (cc @adriancole), I am not sure if we break existing users by this (I guess not). Just one question: do we need `readonly annotationType: string;` in the last case?

jcchavezs: I think the last one is enough (cc @adriancole), I am not sure if we break existing users by this (I guess not). Just one question: do we need `readonly annotationType: string;` in the last case?

jcchavezs: I think the last one is enough (cc @adriancole), I am not sure if we break existing users by this (I guess not). Just one question: do we need `readonly annotationType: string;` in the last case?

jcchavezs: I think the last one is enough (cc @adriancole), I am not sure if we break existing users by this (I guess not). Just one question: do we need `readonly annotationType: string;` in the last case?

adriancole: I think we constrain the typescript api to string, if the underlying javascript can still pass otherwise. I didn't realize that the typescript will affect unit tests written in javascript..

jcchavezs: I think the last one is enough (cc @adriancole), I am not sure if we break existing users by this (I guess not). Just one question: do we need `readonly annotationType: string;` in the last case?

sonybinhle: @jcchavezs @adriancole I updated the PR so with BinaryAnnotation the constructor will only allow string value.

> I think we constrain the typescript api to string, if the underlying javascript can still pass otherwise. I didn't realize that the typescript will affect unit tests written in javascript..
The tests failed is my tests on type-tests folder, I saw the constructor accept boolean|string|number so I wrote tests to verify those three cases. But now when we only accept string then I also modified the ts tests.

> Just one question: do we need `readonly annotationType: string;` in the last case?
It is mandatory I guess since BinaryAnnotation implement the IAnnotation interface, I tried to remove it but getting errors on compiling. And also, the actual implementation has annotationType property.

sonybinhle: @jcchavezs @adriancole I updated the PR so with BinaryAnnotation the constructor will only allow string value.

> I think we constrain the typescript api to string, if the underlying javascript can still pass otherwise. I didn't realize that the typescript will affect unit tests written in javascript.
The tests failed is my tests on type-tests folder, I saw the constructor accept boolean|string|number, so I wrote tests to verify those three cases. But now when we only accept string then I also modified the ts tests.

> Just one question: do we need `readonly annotationType: string;` in the last case?
It is mandatory I guess since BinaryAnnotation implements the IAnnotation interface, I tried to remove it but getting errors on compiling. And also, the actual implementation has annotationType property.

sonybinhle: @jcchavezs @adriancole I updated the PR so with BinaryAnnotation the constructor will only allow string value.

> I think we constrain the typescript api to string, if the underlying javascript can still pass otherwise. I didn't realize that the typescript will affect unit tests written in javascript.

The tests failed is my tests on type-tests folder, I saw the constructor accept boolean|string|number, so I wrote tests to verify those three cases. But now when we only accept string then I also modified the ts tests.

> Just one question: do we need `readonly annotationType: string;` in the last case?

It is mandatory I guess since BinaryAnnotation implements the IAnnotation interface, I tried to remove it but getting errors on compiling. And also, the actual implementation has annotationType property.

sonybinhle:
@jcchavezs @adriancole I updated the PR so with BinaryAnnotation the constructor will only allow string value.

> I think we constrain the typescript api to string, if the underlying javascript can still pass otherwise. I didn't realize that the typescript will affect unit tests written in javascript.

The tests failed is my tests on type-tests folder, I saw the constructor accept boolean|string|number, so I wrote tests to verify those three cases. But now when we only accept string then I also modified the ts tests.

> Just one question: do we need `readonly annotationType: string;` in the last case?

It is mandatory I guess since BinaryAnnotation implements the IAnnotation interface, I tried to remove it but getting errors on compiling. And also, the actual implementation has annotationType property.

sonybinhle:
@jcchavezs @adriancole I updated the PR so with BinaryAnnotation the
constructor will only allow string value.

> I think we constrain the typescript api to string, if the underlying
javascript can still pass otherwise. I didn't realize that the typescript
will affect unit tests written in javascript.

The tests failed is my tests on type-tests folder, I saw the constructor
accept boolean|string|number, so I wrote tests to verify those three cases.
But now when we only accept string then I also modified the ts tests.

> Just one question: do we need `readonly annotationType: string;` in the
last case?

It is mandatory I guess since BinaryAnnotation implements the IAnnotation
interface, I tried to remove it but getting errors on compiling. And also,
the actual implementation has annotationType property.

jcchavezs:
Awesome work @sonybinhle!

PR #146, opened 2017-11-01
Zipkin to operate in node > v8

pbadenski:
zipkin-js source in version 1.X will be written and published for node > v8.

This was previously discussed with @adriancole @SimenB @eirslett (sorry if I missed someone).

Unblocks https://github.com/openzipkin/zipkin-js/pull/135, https://github.com/openzipkin/zipkin-js/pull/208 and https://github.com/openzipkin/zipkin-js/pull/250

PR #272, opened 2018-10-26
WIP: Rollup build

### LINKED ISSUE #270: Use bundler
Maybe zipkin libs should use a bundler like [rollup](https://rollupjs.org/guide/en). This will make libraries lightier because of tree shaking and dead code elimination. 
Front end applications should have a big improvement because of this.

Relates to #195 (comment)

DavideCarvalho:
I'll make a rollup config so when we build the application, we can have .umd, .ecj and .esm files.
This is especially good if you are using a micro frontend aproach and loading everything with systemjs.
Since rollup does dead code elimination, minification and tree shaking, the lib size will be smaller too.

Close #270


DavideCarvalho:
Okay, we have a little problem with the ci...
Sounds like rollup doesn't work with node 4.x, neither with 5.x, only with node 6+
@jcchavezs can we do something about it?

DavideCarvalho:
Node 4.x is quite old and it's not even on LTS anymore, shouldn't the ci change from 4.x to the latest node version?

jcchavezs:
Same happened with ts-runtime, we enabled tests only for node >4. Regarding killing node 4, that would be a major and for that I would also enable support for node >8. There is already a PR that started that work: https://github.com/openzipkin/zipkin-js/pull/146/files. Are you up for such a thing?

José Carlos

jcchavezs:
All the conditionals for node versions are done in https://github.com/openzipkin/zipkin-js/blob/master/.travis.yml#L9

José Carlos

PR #278, opened 2018-11-12
Bugfix/require network

DavideCarvalho:
Sadly, i asked for you guys to merge, since the import for network was working. Since i was changing stuff on zipkin-js inside of node_modules, i guess I forgot to change inside the project itself, so this PR is to fix this.
Really sorry for the mistake

PR #279, opened 2018-11-12
changing import to network file

DavideCarvalho:
Referencing PR #278


jcchavezs:
Thanks @DavideCarvalho! I am curious why our CI did not fail previously and how we could avoid this to happen again.

PR #280, opened 2018-11-12
Update Endpoint isEmpty

briancavalier:
I recently [ported the zipkin TS defs to Flow](https://github.com/flow-typed/flow-typed/pull/2953).  In the process, I found a couple types that seemed incorrect:

* Endpoint isEmpty's implementation returns a boolean, but the TS def was returning void.
* ~~The usual type of flatMap (i.e. monadic bind) is `(a -> Option b) -> Option a -> Option b`.  The TS types seemed to be identical to `map`, but the current Some flatMap implementation _seems_ to be similar to map + monadic join (using getOrElse as join).  I realize there may be a reason (historical?) for the TS type, but this change seemed right and I included it in this PR.~~ We decided to handle this in another PR ([see comment](https://github.com/openzipkin/zipkin-js/pull/280#issuecomment-439188846))

I included both of these changes in the Flow defs, so hopefully they're acceptable 😄.  Thanks for all your work on zipkin-js.

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -80,7 +80,7 @@ declare namespace zipkin {
 
       map<V>(fn: (value: T) => V): IOption<V>;
       ifPresent<V>(fn: (value: T) => V): IOption<V>;
-      flatMap<V>(fn: (value: T) => V): IOption<V>;
+      flatMap<V>(fn: (value: T) => IOption<V>): IOption<V>;
       getOrElse<V>(fnOrValue: (() => V) | V): T;
       equals(other: IOption<T>): boolean;
       toString(): string;
@@ -91,7 +91,7 @@ declare namespace zipkin {
       present: false;
 
       map<V>(fn: (value: any) => V): INone;
-      flatMap<V>(fn: (value: any) => V): INone;
+      flatMap<V>(fn: (value: any) => IOption<V>): INone;
       equals(other: IOption<any>): boolean;
       toString(): string;
     }
@@ -105,7 +105,7 @@ declare namespace zipkin {
 
       map: <V>(fn: (value: T) => V) => IOption<V>;
       ifPresent: <V>(fn: (value: T) => V) => IOption<V>;
-      flatMap: <V>(fn: (value: T) => V) => IOption<V>;
+      flatMap: <V>(fn: (value: T) => IOption<V>) => IOption<V>;
       getOrElse: () => T;
       equals: (other: IOption<T>) => boolean;
       toString: () => string;
@@ -124,7 +124,7 @@ declare namespace zipkin {
       setIpv4(ipv4: string): void;
       setPort(port: number): void;
 
-      isEmpty(): void;
+      isEmpty(): boolean;
     }
 
     interface Annotation {
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -80,7 +80,7 @@ declare namespace zipkin {
 
       map<V>(fn: (value: T) => V): IOption<V>;
       ifPresent<V>(fn: (value: T) => V): IOption<V>;
-      flatMap<V>(fn: (value: T) => IOption<V>): IOption<V>;
+      flatMap<V>(fn: (value: T) => V): IOption<V>;
       getOrElse<V>(fnOrValue: (() => V) | V): T;
       equals(other: IOption<T>): boolean;
       toString(): string;
@@ -91,7 +91,7 @@ declare namespace zipkin {
       present: false;
 
       map<V>(fn: (value: any) => V): INone;
-      flatMap<V>(fn: (value: any) => IOption<V>): INone;
+      flatMap<V>(fn: (value: any) => V): INone;
       equals(other: IOption<any>): boolean;
       toString(): string;
     }
@@ -105,7 +105,7 @@ declare namespace zipkin {
 
       map: <V>(fn: (value: T) => V) => IOption<V>;
       ifPresent: <V>(fn: (value: T) => V) => IOption<V>;
-      flatMap: <V>(fn: (value: T) => IOption<V>) => IOption<V>;
+      flatMap: <V>(fn: (value: T) => V) => IOption<V>;
       getOrElse: () => T;
       equals: (other: IOption<T>) => boolean;
       toString: () => string;


jcchavezs:
Awesome work! I think we can merge this by tomorrow. Pinging @sonybinhle as
he is our TS master.

Den man. 12. nov. 2018, 22:16 skrev Brian Cavalier <notifications@github.com
:

sonybinhle:
@briancavalier Thank you very much for pointing out the differences between ts definition vs the actual implementation. I wrote some tests for flatMap ts definition and my result is a little bit different with yours. 
https://github.com/openzipkin/zipkin-js/compare/master...sonybinhle:feature/test-type-option?expand=1

briancavalier:
Thanks for the feedback, @sonybinhle.

It sounds like flatMap is not intended to be the categorical monadic bind, and that it's best to make the TS types match whatever the implementation is doing now.  I should have time today to do that.

However, I think it's worth noting that the name "flatMap" will tend to create expectations about its behavior, as it did for me.  For example, anyone familiar with Option or Maybe types from Scala, Haskell, Purescript, JS fantasy-land, etc. and/or with the name "flatMap" from other contexts will likely expect it to be monadic bind and will find it surprising that it isn't.

On the other hand, zipkin-js's Option implementation doesn't claim to be a monad, so I think it's ultimately fine to have a method which happens to be named "flatMap" and does whatever the zipkin-js team wants/needs it to do (with the caveat about expectations).

packages/zipkin/index.d.ts
```diff
@@ -80,7 +80,7 @@ declare namespace zipkin {
 
       map<V>(fn: (value: T) => V): IOption<V>;
       ifPresent<V>(fn: (value: T) => V): IOption<V>;
```
briancavalier: I just noticed that [None's implementation of ifPresent always returns undefined](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/option.js#L8) rather than IOption.  Would you like to deal with that in this PR?  If so, let me know your thoughts on how to proceed.  Thanks!

packages/zipkin/index.d.ts
```diff
@@ -80,7 +80,7 @@ declare namespace zipkin {
...
-      flatMap<V>(fn: (value: T) => V): IOption<V>;
+      flatMap<V>(fn: (value: T) => IOption<V>): IOption<V>;
```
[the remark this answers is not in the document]
briancavalier: Thanks. I'm totally fine with making the types match the reality of the implementation.

One observation here is that returning a union makes this a bit more cumbersome for callers.  If all they know is that they have an IOption, they'll have to check flatMap's return value.  For example, TS users will be prevented from chaining any IOption methods after flatMap: `anOption.flatMap(f).map(g)` will be disallowed.

packages/zipkin/index.d.ts
```diff
@@ -80,7 +80,7 @@ declare namespace zipkin {
...
       getOrElse<V>(fnOrValue: (() => V) | V): T;
```
briancavalier: Hmmm, I think getOrElse's type may be incorrect as well.  For example, None can't possibly return a T, since None represents the absence of a T.  I see two possible ways to go:

1. `getOrElse(fnOrValue: (() => T) | T): T;`  Personally I think this more rigorous type is preferable, as it makes the strong guarantee to the caller that he/she will definitely get a T: Some can return the T it already has, and None can return a T via the argument.
2. `getOrElse<V>(fnOrValue: (() => V | V): T | V;`  TS callers will be forced (and JS callers should be advised) to test the return value to know whether they got a T or a V.

The current Some and None implementations are compatible with both of those.  Let me know what you think.

packages/zipkin/index.d.ts
briancavalier: Looking at usages of ifPresent in the zipkin-js codebase, it seems like the return value is always ignored.  That makes it seem like we could change the return type to `void`.  Thoughts?

jcchavezs:
Well if it is part of the public API (i.e. someone could use it) then we
better keep it as non void.

Den tor. 15. nov. 2018, 20:03 skrev Brian Cavalier <notifications@github.com
:

briancavalier:
@jcchavezs I hear you.  However, as it stands, this is a bug: the type of ifPresent and the None implementation don't agree.  If a TS user has an IOption in their hands, and it turns out to be None at runtime, ifPresent will return undefined and their code will break at runtime.

Changing the return type to `void` is only one option, of course.  Another is to change the implementation of None ifPresent to return IOption, e.g. it can return None.

packages/zipkin/index.d.ts
eirslett: `ifPresent` was never intended to return a value. It's purely a way of performing side effects. The correct signature is indeed `void`.

eirslett: Update: Yup, there is actually a bug in the implementation. `Some` returns a value for `ifPresent`, but it shouldn't. I guess the type signature matches the implementation, so the bug is both in the implementation and the signature.

briancavalier:
@eirslett Thanks for the context!  Going with TS + nullables sounds like a good direction.

In the short term, can you help me understand what that may mean for this PR?  For example, should we do anything with the flatMap type?  My original changes match the [Scala Option flatMap](https://www.scala-lang.org/api/current/scala/Option.html#flatMap[B](f:A=>Option[B]):Option[B]), but it seems like maybe [the current Some implementation doesn't](https://github.com/openzipkin/zipkin-js/compare/master...sonybinhle:feature/test-type-option?expand=1#diff-75188cb7943d6025abff1ed019dfdc29R39).

Again, I'm happy to help in whatever way the zipkin-js team feels is best.

eirslett:
In the short term, I think the best way forward is to fix any bugs in Option/Some/None, which means they should have the same semantics and behaviour as in Scala. And also fix the type definitions accordingly.

These types were always meant for internal use by zipkin-js, I don't think we expose them to consumers of the library? (Maybe I don't remember correctly...)

jcchavezs:
Agree, I just checked packages and none seem to use them. Changing them
should not be a problem.

Den tor. 15. nov. 2018, 20:39 skrev Eirik Sletteberg <
notifications@github.com:

briancavalier:
@eirslett @jcchavezs Ok, thanks.  @sonybinhle does that sound reasonable to you as well?

Do you envision doing that in a separate PR?  For example, I could reduce this PR to just the isEmpty boolean change.  I don't mind helping with a larger set of changes to Option, but I may be fairly slow.

packages/zipkin/index.d.ts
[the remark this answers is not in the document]
sonybinhle: I totally agree with you about the cumbersome and mess public API. As you said below, there are two ways of solving it, either changing TS definition or changing actual implementation. But changing the actual implementation is hard because we don't know the effects to functions using it. I also favor changing the actual implementation if it is not consistent but I would ask @jcchavezs for a recommendation.

packages/zipkin/index.d.ts
sonybinhle: I also like the first one more, I think it is the purpose of an interface to provide the user with a common rigid signature otherwise the user may tend to implement the wrong abstraction.

eirslett:
To be honest, I don't have much time to follow up on this, or get more involved, but I think @jcchavezs is already a bit involved in the PR?

jcchavezs:
@briancavalier @sonybinhle I think it is easier to keep PRs small so if you could address it in a different PR that would be awesome. Really thanks for your collaboration on this and @eirslett thanks for jump in!

briancavalier:
Sounds good, @jcchavezs.  I'll work on reducing this PR first, and then on another that focuses on Option's type and implementation.

I appreciate all the help so far!

briancavalier:
Ok, this PR now only includes the isEmpty change.

jcchavezs:
Great @briancavalier! Any input @sonybinhle?

sonybinhle:
> Any input @sonybinhle?
I verified the actual function, should it be enough for this case?

briancavalier:
🎉

PR #281, opened 2018-11-13
Documentation

DavideCarvalho:
I started the documentation with docusaurus, so I'd like your thoughts about the tool, or if you guys have any other idea about how to document it.
To start the documentation, go to website folder and npm start it.
One more thing, if you guys want to create the documentation when the code runs on travis, it needs node 8.2 or above.


DavideCarvalho:
This branch is on "WIP" because I think it needs more stuff before releasing it as at least an initial documentation.
I'd like to know what it needs to be done before when can release this documentation.

devinsba:
/cc @abesto if you don't have a watch in this repo

I think we might want to consider a long term strategy here with the move into the ASF incubator, it might do us good to have the language docs published to a centralized place, even if they are built in the individual repos

DavideCarvalho:
@devinsba according @adriancole, this project won't be to ASF at this moment

adriancole:
my 2p: yep technically this isn't going into the ASF in the current effort. this doesn't mean it doesn't later, nor does it imply the people are exactly the same regardless :D https://wiki.apache.org/incubator/ZipkinProposal

I would think about @devinsba comment would be less about ASF in general and more about the community. Does it or does it not serve us to put docs in a centralised place? Right now, we have some hybrid docs, for common areas of troubleshooting on the wiki. If nothing else, related concerns should link to that place IMHO Edit access to wiki for example does not require a formal ASF account (even if it does need an arbitrary user created) https://cwiki.apache.org/confluence/display/ZIPKIN

My opinion I think that it is good to have Brian's note in the back of mind, but don't get too distracted about a potential future of doc aggregation. Reason is that there's the typescript project etc and while things are in flux maybe best focus on the tasks at hand.

DavideCarvalho:
I removed some files and added a folder structure that I think it's the best for us to keep the documentation.
I guess will be good if we have some CONTRIBUTING.md so we can say how people can help both with the documentation and the code

adriancole:
might be under .github

see also https://github.com/openzipkin/zipkin/tree/master/.github which has
been changed more recently

jcchavezs:

PR #287, opened 2018-11-16
Droping node 4 support

DavideCarvalho:
Removed node4 from travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,6 @@ language: node_js
 
 matrix:
   include:
-    - node_js: 4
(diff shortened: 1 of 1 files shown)


DavideCarvalho:
The README says that Node 6 or newer is required, so we already have a minimum version

DavideCarvalho:
I guess we need to put on README that this library is intented for browsers as well (In clock precision session, we say about browsers, but we need to say in a more explicite way)

adriancole:
👍  thanks!

Issue #283, opened 2018-11-13
Port the whole zipkin-js codebase to TypeScript


eirslett:
This is a suggestion.

For the v1.0 milestone, we could port the whole zipkin-js codebase to TypeScript. We could then publish the generated type bindings.
It would take some work to do the actual rewriting, but I'm sure somebody would be interested in doing the actual work (If not, then of course it's not feasible).

The gains would be:
- Type-checking the whole codebase at build time can potentially find bugs
- We don't have to maintain separate typings for TypeScript, that can potentially be out of sync

Problems:
- It's no longer possible to run the zipkin-js codebase "as-is" in node.js, it has to be transpiled with a TypeScript workflow
- People who don't know how to work with TypeScript might get a higher barrier to entry
- Some developers just don't like TypeScript

DavideCarvalho:
Nice issue, but i’d like to ask you: Why Typescript and not Flow?

DavideCarvalho:
My thoughts:
- Transpile the zipkin codebase shouldn't be a problem, since it already uses Babel, if we can upgrade to babel 7, we can use @babel/preset-typescript for transpiling TS
- TS isn't really that hard, and the only thing it adds its type checking, this shouldn't be a problem.

I guess we don't have to consider people disliking TS, but think that a type checking system will ensure that the code is more maintainable and less error prone. What we have to consider is if TS can solve what we need, or if we have to go to another solution, like Flow.

eirslett:
My impression is that Flow is somewhat better for React frontend projects and TypeScript is better for backend/node.js projects, but that's not _really_ the case. For every release, they seem to get more similar. They are practically interchangeable. Flow can do feature X and TypeScript can do Y etc (IDE integration, gradual typing, features of the type system, IDLs for common APIs and libraries...)., but for every release they seem to catch up with each other.

Traction-wise, it looks like TypeScript is being more used than Flow. That's the only good reason I can come up with.

DavideCarvalho:
From my experience with Flow, TS ecosystem is much more developed and stable, so I agree that TS is the best choice.
I think that people that dislike TS it's because TS is a super set of JavaScript, but with Babel TS, it becomes just a type checker, just like Flow, so I guess people will dislike it less.

jcchavezs:
Thanks @eirslett for bringing this over the table. This sounds like a great idea however I have some concerns:

[...]

PR #290, opened 2018-11-21
TypeScript Port

DavideCarvalho:
Starting to port zipkin codebase from JavaScript to TypeScript.

Must do:
- [ ] Port all zipkin packages
- [ ] Fix TSLint errors (maybe we should see what rules make sense to us and disable what we don't want)
- [ ] Fix tests errors


DavideCarvalho:
Referencing #283

DavideCarvalho:
What I could see from the tests i've made is that the build is ok, but the tests are breaking because ```require``` resolutes only to ```.js``` files. In order to get ts files, we need to explicitly tell the extension in the end, e.g ```require('./INetAddress.ts')```.
We can try to change to ```import``` and ```export```. Either way, we'll need to refactor the code.

eirslett:
Looks like a good step is to rename all `.js` files to `.ts`, and run TypeScript with a loose/liberal config - then we can tighten it once more files are typed correctly?

Maybe we could run tests with ts-node? https://github.com/TypeStrong/ts-node

DavideCarvalho:
The tests are running well with mocha and babel, I don't think ts-node will be needed. The only problem is with requiring files, since require tries to find .js files by default, and to require .ts files we need to put the extension along with the file name.

DavideCarvalho:
When I get home I will change the require statements and test it with zipkin package.

DavideCarvalho:
Strange, locally the tests are passing, while on CI it's getting ```../../node_modules/@types/node/index.d.ts(7265,88): error TS2304: Cannot find name 'Symbol'.```.

eirslett:
> trange, locally the tests are passing, while on CI it's getting ../../node_modules/@types/node/index.d.ts(7265,88): error TS2304: Cannot find name 'Symbol'..

Old node.js version?

DavideCarvalho:
I was using node 11, I changed to 6 and it still passes locally.

jcchavezs:
Weird, anything you see @sonybinhle?

Den ons. 21. nov. 2018, 16:12 skrev Davi de Carvalho <
notifications@github.com:

jcchavezs:
Thanks @sonybinhle if that is the case, can we do that in a separated PR so
we integrate the change earlier? What do you think @DavideCarvalho?

José Carlos Chávez


Den tor. 22. nov. 2018 kl. 15:24 skrev Binh Le <notifications@github.com>:

sonybinhle:
> if that is the case, can we do that in a separated PR so we integrate the change earlier

Another way of solving it is by removing `"typescript": "^2.4.2"` from `packages/zipkin/package.json`. The global one is `"typescript": "^2.9.2"` seems to work fine with new `tsconfig.json`.

Another observation is that `tsc index.d.ts` is only be included in `packages/zipkin/package.json` but not on the other sub-packages. So maybe our purpose here is checking the packages/zipkin/index.d.ts` is valid with typescript version 2.4.2 

I think we should remove local typescript version and upgrade global typescript if it doesn't hurt any usages.

eirslett:
Let's just go for latest TypeScript (v3.1) everywhere. There's no need for backwards compatibility since this will be v1.0.

DavideCarvalho:
Oh, I have TS installed globally, maybe it's because of this it was working on my machine. Okay, i'll update TS version and try again

jcchavezs:
@DavideCarvalho we merged a couple of PRs that are going to be part of the last 1.0. Could you please rebase your PR accordingly?

DavideCarvalho:
```Error: Cannot find module './option.ts'```, I tried looking finding where is calling 'options.ts' but I couldn't debug it to the root of the problem. Something at ```packages/zipkin-instrumentation-cujojs-rest/test/integrationTest.js``` is calling a option.ts file.

First I thought that inside zipkin's lib folder was something calling a ```options.ts``` file, but I couldn't find.

jcchavezs:
I see a pattern here and this is totally on me. I should have foreseen this before.

Zipkin JS is getting a lot of attraction these days and we get [more and more PRs every day](https://github.com/openzipkin/zipkin-js/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+sort%3Acreated-desc+language%3Ajavascript+merged%3A%3E2018-11-01+) (13 merged the last two weeks).

This PR has been created 6 days ago and since then we merged 8 PRs, every time we merge a new PR @DavideCarvalho has to do a rebase and with that there is a risk new stuff come up. This is definitively not desired, efficient nor fair and it can get him burned out.

On the other hand, we can not just merge keep those PRs hanging until we finish this titanic work (which I will be in debt with @DavideCarvalho and all you guys for a while) as this is mainly a volunteer project and 4 months ago it was almost dead so we want contributors to be proactive and open PRs to solve issues but we also want to fulfill our part of the treat by getting stuff merged and released ASAP.

In that direction I see the future of this PR (if we keep the same behaviour) becoming so complex and error prone because every time the scope of the PR becomes endless that it is in risk of not getting to the goal.

My proposal is:
1. We instead of moving JS files into TS files, do that but into a folder called `packages-ts` but we test both on CI.
2. Every time someone a new PR we request them to do the counter part in packages-ts (being a blocking rule)
3. We keep releasing 0.x versions of zipkin but don't make `packages-ts` official.
4. Once the move to TS is done we just drop the JS (which should be the same).

This approach has many advantages:
1. We can move one by one the packages independently (smaller PRs merged more often)
2. Every change in js files have a counter part in TS done by the contributors.
3. We don't block @DavideCarvalho anylonger.
4. Someone else can help him.

What do you think peeps? @openzipkin/zipkin-js-champions @eirslett @ghermeto 

Any other idea that help us to integrate this changes earlier without breaking existing clients will be also considered.

eirslett:
I suggest a "simple" solution:

We rename every .js file to .ts, set up a build script (noop-ing, just using babel removing type annotations) and then we merge that PR right away.
Remember, all JavaScript is still valid TypeScript, so it won't break anything.

There won't be any merge conflicts or need to do double work.

The technique is called gradual typing, and means that the types are opt-in instead of mandatory.

We configure the TypeScript validator to be very liberal (we might even simply disable it for now!)

Moving forward, we must make sure all TypeScript-related branches are very short-lived, so they don't collide with other incoming pull requests. Therefore, every TypeScript-PR must also be kept minimal, and there must be emphasis on reviewing and merging them quickly (time-wise).

DavideCarvalho:
I agree with the gradual typing, and it’s what I’m aiming with this PR, but I’m getting some errors on the ci and are errors that I’m not getting locally, don’t know why. If someone could test it locally as well, I could make the change and see if it passes. I’lol try again at home hoping to catch those errors to debug

ghermeto:
@jcchavezs this is a classical migration issue. Usually the solution is to focus on small verticals. Fortunately `zipkin-js` provides a easy and unique opportunity for that: it is all split into smaller packages

You don't need to add typing is all packages or even rename files in all packages. You can start with renaming the packages on the most important package: `zipkin` and from there hit the transports, context and later the interceptors.

Honestly I don't even see a reason for the packages that will run only in `node.js` to be transpiled at all (IMO it is only good to make production debugging a nightmare). Of course you want some consistency, but that can be eventually achieved (and if makes sense)...

For that to work though, you can't wait until `1.0`. You have to merge the PRs as they arrive.

jcchavezs:
@ghermeto thanks for the insight. @DavideCarvalho maybe @sonybinhle can help you?

sonybinhle:
@DavideCarvalho After the build, the bundle is in the **lib** folder. `"@babel/preset-typescript"` transpile `const option = require('./option.ts');` to `var option = require('./option.ts');` instead of `var option = require('./option');`.

We could replace all required `.ts` extensions with **empty** after building by a `script` or we could find any existing babel plugins to do that job for us. The alternative solution could be removing all `.ts` required extension from typescript files since having `require('xxxx**.ts**')` may be unnecessary.

DavideCarvalho:
I find it very strange, since nothing inside ```lib``` folder is calling for a ```.ts``` file.

sonybinhle:
@DavideCarvalho Maybe you have some global config that removes `.ts` extension. The result if I run build in my computer.

screen shot 2018-11-28 at 11 31 22 am

DavideCarvalho:
Okay, I guess this branch is just too far behind and is quite hard to rebase, so I will create another branch to finish this task.
Instead of doing ```require(file.ts)```, looks like ```import``` and ```export``` statements are transpiled without the file extension, so I guess using them will be easier, I'll do it on zipkin package first.

DavideCarvalho:
I changed two imports on index.ts file and it worked! Build and tests are running, I'll push it here to see how the ci will behave

jcchavezs:

PR #294, opened 2018-11-22
adds configurable logging module as argument and defaults to console

ghermeto:
Enables the application to pass a logging module as argument to be used instead of hardcoding to `console`, which will just print a string.

This is specially important for the transports, as production applications will likely use logging aggregation which will expect a specific format.

I was on the fence about updating the method `writeIdToConsole` in the `Tracer` class, but I did it for consistency.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 node_modules
 gen-nodejs
 lib
+**/test-types/*.js
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ yarn-error.log
 packages/**/package-lock.json
 **/test-types/*.test.js
 **/tsr-declarations.js
+.idea/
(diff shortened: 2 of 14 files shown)


jcchavezs:
Thanks @ghermeto <https://github.com/ghermeto> I really love this change
however my experience is that such failures can flood logs in production,
what if we provide a noop by default instead of console? Otherwise we can
just merge this as it keeps the current behaviour. Any input @adriancole?

Den tor. 22. nov. 2018, 01:10 skrev Guilherme Hermeto <
notifications@github.com:

ghermeto:
I just realized this PR needs more work. I need to update the TypeScript type definitions and document the option in the README files.

I can see both ways regarding defaulting to a `noop` or `console`. The `HttpLogger` will aggregate the traces for one second (by default), so it isn't so bad. However, the `KafkaLogger` will immediately publish, which may indeed flood the logs...

ghermeto:
@jcchavezs I thought more about having a default noop logging, but I think we should keep the current behavior for now (using `console`). 

The reasoning is that with this new option no one should be using `console` in production and having the errors showing up on `stderr` is helpful for early development.

I added an example on the README files how to disable the `console` by using `noop-logger`.
Hopefully that is enough...

jcchavezs:
@ghermeto makes sense. I will merge this as it is. Thanks @ghermeto!

PR #296, opened 2018-11-22
add timestamp parameter to recordAnnotation.

haoguanjun:


ghermeto:
can you also update the TypeScript definitions on `packages/zipkin/index.d.ts`?

packages/zipkin/index.d.ts
```diff
@@ -39,7 +39,7 @@ declare namespace zipkin {
     createChildId(): TraceId;
     letId<V>(traceId: TraceId, callback: () => V): V;
     setId(traceId: TraceId): void;
-    recordAnnotation(annotation: IAnnotation): void;
+    recordAnnotation(annotation: IAnnotation, timestamp?: number): void;
```
jcchavezs: Is this number type OK? Shouldn't we restrict to integers?

jcchavezs:
Thanks *@DavideCarvalho*!

José Carlos Chávez


Den fre. 23. nov. 2018 kl. 11:24 skrev Davi de Carvalho <
notifications@github.com>:

jcchavezs:
A second though came to my mind about what @eirslett pointed out. We are
not explicit about what units are being used here. What about changing
timestamp into timeInMicros or something like that? CC @adriancole

Den fre. 23. nov. 2018, 12:44 skrev José Carlos Chávez <jcchavezs@gmail.com:

adriancole:
the question about units is frequent and people often make mistakes. it is
tempting to use a normal unit, but then there are awkward and useless ones
usually like hours or days. recording with pre-ordained timestamp is hairy
and likely to cause bugs. I'd rather hazard in the form of docs vs types.
IOTW microseconds. That's my usual answer but it isn't what everyone thinks
:)

>

PR #297, opened 2018-11-23
small performance improvements for restify middleware

ghermeto:
1) using `bind` is slower compared with other methods:

The following test:
```javascript
var Benchmark = require('benchmark');
var suite = new Benchmark.Suite;

class Some {
    constructor(val) {
        this.val = val;
    }
}

class None {}

const req = {
    headers: { foo: 'bar' },
    header(name) {
        return this.headers[name];
    }
};

function headerOption(request, header) {
    const val = request.header(header);
    if (val != null) {
        return new Some(val);
    } else {
        return None;
    }
}

suite.
    add('redefining', function() {
        function readHeader(header) {
            const val = req.header(header);
            if (val != null) {
                return new Some(val);
            } else {
                return None;
            }
        }
        return readHeader('foo');
    }).
    add('binding', function() {
        const readHeader = headerOption.bind(null, req);
        return readHeader('foo');
    }).
    add('wrapping', function() {
        const readHeader = (header) => headerOption(req, header);
        return readHeader('foo');
    }).
    on('cycle', function(event) {
        console.log(String(event.target));
    }).
    on('complete', function() {
        console.log('Fastest is ' + this.filter('fastest').map('name'));
    }).
    run();
```

produce the results:
```
redefining x 120,611,860 ops/sec ±1.59% (84 runs sampled)
binding x 37,074,641 ops/sec ±1.00% (82 runs sampled)
wrapping x 137,376,097 ops/sec ±1.47% (85 runs sampled)
Fastest is wrapping
```

2) `instrumentation` can be instantiated before the middleware:

this change moves the instantiation to the wrapper function in the same way as the `express` middleware.

### DIFF
--- a/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
+++ b/packages/zipkin-instrumentation-restify/src/restifyMiddleware.js
@@ -1,7 +1,4 @@
-const {
-  option: {Some, None},
-  Instrumentation
-} = require('zipkin');
+const {option: {Some, None}, Instrumentation} = require('zipkin');
 const url = require('url');
 
 function headerOption(req, header) {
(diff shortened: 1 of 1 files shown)


jcchavezs:
This is wonderful. Re-runing the tests.

jcchavezs:
This is green. Amazing @ghermeto!

PR #298, opened 2018-11-23
Support creating a new spanId on HttpServer instrumentation.

juhovh:
Some tracing systems like Amazon X-Ray, Google Stackdriver Trace, LightStep or OpenTracing specification require each span to have a unique spanId, but currently using zipkin-js for tracing to those systems breaks the traces.

This change modifies the HttpServer instrumentation so that it is possible to force creation of a new id after reading the headers. It doesn't add support to other packages like express or hapi, but makes it easy to write a new server instrumentation with unique spanId support.

### DIFF
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -31,17 +31,19 @@ class HttpServerInstrumentation {
     serviceName = tracer.localEndpoint.serviceName,
     host,
     port = requiredArg('port'),
+    supportsJoin = true,
   }) {
     this.tracer = tracer;
     this.serviceName = serviceName;
     this.host = host && new InetAddress(host);
     this.port = port;
+    this.supportsJoin = supportsJoin;
   }
 
   _createIdFromHeaders(readHeader) {
     if (containsRequiredHeaders(readHeader)) {
       const spanId = readHeader(Header.SpanId);
-      return spanId.map(sid => {
+      const parentId = spanId.map(sid => {
         const traceId = readHeader(Header.TraceId);
         const parentSpanId = readHeader(Header.ParentSpanId);
         const sampled = readHeader(Header.Sampled);
@@ -54,6 +56,13 @@ class HttpServerInstrumentation {
           flags
         });
       });
+      return !this.supportsJoin
+        ? parentId.map(id =>
+            this.tracer.letId(id, () =>
+              this.tracer.createChildId()
+            )
+          )
+        : parentId;
     } else {
       if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
         const sampled = readHeader(Header.Sampled) === None ?
--- a/packages/zipkin/test/httpServerInstrumentation.test.js
+++ b/packages/zipkin/test/httpServerInstrumentation.test.js
@@ -118,6 +118,63 @@ describe('Http Server Instrumentation', () => {
     });
   });
 
+  traceContextCases.forEach((headers, index) => {
+    it(`should should not join spans if join not supported case ${index}`, () => {
+      const record = sinon.spy();
+      const recorder = {record};
+      const ctxImpl = new ExplicitContext();
+      const tracer = new Tracer({recorder, ctxImpl});
+
+      const port = 80;
+      const host = '127.0.0.1';
+      const url = `http://${host}:${port}`;
+      const instrumentation = new HttpServer({
+        tracer,
+        serviceName: 'service-a',
+        port,
+        supportsJoin: false
+      });
+
+      const readHeader = function(name) {
+        return headers[name] ? new Some(headers[name]) : None;
+      };
+      ctxImpl.scoped(() => {
+        const id = instrumentation.recordRequest('POST', url, readHeader);
+        tracer.recordBinary('message', 'hello from within app');
+        instrumentation.recordResponse(id, 202);
+      });
+      const annotations = record.args.map(args => args[0]);
+
+      annotations.forEach(ann => expect(ann.traceId.traceId).to.equal('aaa'));
+      annotations.forEach(ann => expect(ann.traceId.parentId).to.equal('bbb'));
+      annotations.forEach(ann => expect(ann.traceId.spanId).to.not.equal('bbb'));
+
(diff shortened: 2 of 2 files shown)


jcchavezs:
@juhovh thanks for coming by. I am not entirely sure if I understand the problem, is it that zipkin is using same spanID for both client and server and you want them to be different? If that is the case we have the possibility to enable and disable this for other languages (e.g. zipkin-go) under the concept of `shared` span. In such cases we handle with this option at [tracer level](https://github.com/openzipkin/zipkin-go/blob/master/tracer.go#L102). Any input @adriancole?

juhovh:
@jcchavezs Thank you for your fast reply. Yes, that is basically the problem, I want the spanIds to be different for client and server.

I would definitely prefer to handle this on the tracer level, but the way zipkin-js is structured even if this option would be handled on the tracer level some changes to the HttpServer would still be necessary. That's because e.g. expressMiddleware instrumentation takes only tracer as the argument and then creates a Instrumentation.HttpServer instance internally. If you look at the HttpServer instrumentation closely, you notice that `_createIdFromHeaders` makes the decision to either return the client span details or a new root span, meaning creating a child after the call would be too late. Otherwise you would also create a child for the root spans, which would result in parent span not found in the recorded results.

So basically to do this on the tracer level there are two options:
1. query the `supportsJoin` flag of the tracer in the middleware and pass it down to HttpServer
2. move the root span generation logic from HttpServer to tracer as in zipkin-go as it should

I went with the smallest change in order to not break any existing instrumentations out there, because the "proper" way of doing the option 2 above might break many 3rd party instrumentations. Currently zipkin-js library doesn't support server span customization either so in my current codebase I've had to fork the express middleware internally as well and have no problem passing the `supportsJoin` flag to HttpServer, so I'm working around zipkin-js limitations in multiple ways here.

I've worked with the latest Brave Java API a fair bit and consider it to be very mature, it is using the term `supportsJoin` which I used here as well, could've also use the term shared.

adriancole:
yes I believe this addressed primarily at tracer level as other RPC will
have the same concern, not just http. In fact join only works when a span
ID was extracted. here's the logic which handles the routine concern about
extraction but also doesn't place the flag for supportsJoin inside the http
abstraction
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpServerHandler.java#L83

Thanks for thinking about the context.

>

juhovh:
For Brave I think it's even a bit more complicated than that, because the default value for supportsJoin is actually stored in Propagation, where the default value is `false` but it is set as `true` for B3Propagation because of compatibility with the Zipkin behavior. This can be overridden in the Tracing builder, which will then be passed on to the current tracer.

Here in zipkin-js there's much less abstractions and HttpServer is basically doing implicitly both B3Propagation and Sender functionality. So the consensus here is that I should move this for tracer, is it ok if I add a `nextSpan` logic to the tracer and call that from HttpServer instead? That would mean I need to create a similar abstraction as `TraceContextOrSamplingFlags` on the Brave side (basically a context with no guaranties of a spanId being available).

adriancole:
@juhovh I may not completely understand. If your goal is to ensure that you never share a trace ID, having joinSpan or similar on the tracer.. how would that not work? the tracer gets the last say. That keeps a single configuration item regardless of propagation type and regardless of that B3 happens to be the only propagation type in zipkin-js right now.

Ex in the component that configures the tracer, it is a double-and, which means tracer wins on false regardless.
```java
          builder.supportsJoin && propagationFactory.supportsJoin(),
```

here's where it is used there.
```java
  public final Span joinSpan(TraceContext context) {
    if (context == null) throw new NullPointerException("context == null");
    if (!supportsJoin) return newChild(context);
```
https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/Tracer.java#L172

In other words, let's please indeed solve it, but not require this to be thought of at the http abstraction. The config override should be at the tracer layer.

So for example in the tracer, we currently have `createChildId`
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/tracer/index.js#L77

but in the http layer we manually create an id or make a new root id
```javascript
        return new TraceId({
          traceId,
          parentId: parentSpanId,
          spanId: sid,
          sampled: sampled.map(stringToBoolean),
          flags
        });
```

What I'm suggesting is instead of using `tracer.setId` like this..
```
this._createIdFromHeaders(readHeader).ifPresent(id => this.tracer.setId(id));
```

make something like `tracer.joinId` which will join if supported or make a new child if not. IOTW, let's keep the important logic in the tracer and ignore that there is not yet a way to secondarily control via custom http propagation plugin.

I don't think `nextSpan` as defined in brave should be applied here as zipkin-js is still very unlike brave (and zipkin-php and zipkin-go). hopefully, we will change that later, but I think the minimum you need is an operator which abstracts the mechanics of joinSpan (ex `tracer.setId` vs `tracer.createChildId()`. The rest of the stuff can be left alone and might better be left alone until a decision on api rewrite is made.

Make sense?

adriancole:
I think I understand what you are saying and thanks for clarifying it. I guess you end up needing the nextSpan logic as the partial extraction case is what you are highlighting. I'd be happy to review as I think you understand the points I was concerned about.

Meanwhile, I do think it is high time to consider doing the port as that very much simplifies discussions. Unfortunately I don't have time to offer to doing that port :) but it is a lot easier to reason with code that behaves abstractly the same. For example, we are doing translations for no other accident than history. I think this library could do well to "v2-ify" since it will be incompatible from the typescript move anyway. 


Regardless.. my minimum request is to source this setting supportsJoin from the tracer. I really don't want to carry the burden of explaining why this is set in the http layer.

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -31,17 +31,19 @@ class HttpServerInstrumentation {
     serviceName = tracer.localEndpoint.serviceName,
     host,
     port = requiredArg('port'),
+    supportsJoin = true,
```
adriancole: minimum request is to have this set on the tracer and have the httpServer inherit it. I know it is only used here, but it will help us from establishing a pattern which is at the wrong abstraction

juhovh:
I have updated the PR based on the feedback and moved the flag to the tracer. Currently it's only used by HttpServer, but then again spans are usually only joined on B3 RPC propagation.

PR #299, opened 2018-11-23
Add support for maxPayloadSize option in HttpLogger.

juhovh:
Our problem is that having interval as the only way to control the data sent in HttpLogger is not acceptable for two reasons:
1. We have a limitation that our Zipkin server endpoint can only accept payloads up to a certain size
2. We want to reduce memory consumption of zipkin-js in cases where very large amounts of spans are sent in a short timeframe, which means we need to trigger a send earlier than the interval would otherwise trigger it

The solution is to add an option `maxPayloadSize` to HttpLogger, which will ensure that the send payload will never be larger than the set value in bytes. In case the queue plus the added span would go over the limit, the queue is flushed before the span is added to it. In case a single span itself would go over the limit, then the queue is flushed and the large span is dropped.

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -19,13 +19,16 @@ class HttpLogger extends EventEmitter {
     httpInterval = 1000,
     jsonEncoder = JSON_V1,
     timeout = 0,
+    maxPayloadSize = 0,
     /* eslint-disable no-console */
     log = console
   }) {
     super(); // must be before any reference to *this*
     this.log = log;
     this.endpoint = endpoint;
+    this.maxPayloadSize = maxPayloadSize;
     this.queue = [];
+    this.queueBytes = 0;
     this.jsonEncoder = jsonEncoder;
 
     this.errorListenerSet = false;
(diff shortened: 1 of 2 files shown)


adriancole:
good idea.. note that max payload size can be even more concerning in kafka. This doesn't mean your work is not correct, just saying the problem is generic

juhovh:
@adriancole You are right, I just don't have a test setup for kafka myself so I am a bit hesitant to make a change there purely based on the integration tests. I think it probably needs to be done for each transport separately though, I quite like the current simplicity of the reporter interface (only `logSpan` method used).

adriancole:
there is a project zipkin-reporter-java which abstracts the sizing function. it is indeed a little more complex for some use cases but since Kafka and http accept literally same messages (json span list) can be considered. main thing I am concerned about is api drift. we should just make sure the controls are the same even if this is implemented separately.

adriancole:
here is that code btw if using same parameter names should be good I think
https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin2/reporter/Sender.java

jcchavezs:
This looks great to me, thanks @juhovh. My only concern is that we are not
testing the case "In case the queue plus the added span would go over the
limit, the queue is flushed before the span is added to it". Am I wrong?

Den lør. 24. nov. 2018, 01:57 skrev Adrian Cole <notifications@github.com:

> here is that code btw if using same parameter names should be good I think
>
> https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin2/reporter/Sender.java
>

juhovh:
@jcchavezs You are absolutely right, I have added a test case for that now and also fixed a bug in the error emitting test.

Extra points for using a Danish/Norwegian email client. 😄

jcchavezs:
Thanks for the great work @juhovh!

PS. It is norwegian :p

PR #300, opened 2018-11-27
adding instrumentation for connect (works for both express and restify)

ghermeto:
Reasoning: when building frameworks that will augment either Express or Restify, it is preferable to have an agnostic way, which leverage the common methods and properties available on `http.IncomingMessage` and `http.ServerResponse` instead of server specific methods/properties.

Added integration tests for Restify 7.2, Express 4.13 and Connect 3.6

@jcchavezs @adriancole I understand that is something you might not want to merge on this project, as you already have separate instrumentation for Express and Restify, but I assume my use-case isn't that uncommon. I can always publish this on it's own module... let me know.


adriancole:
In brave we do same thing. ex we have http abstraction and also servlet for
things that use servlet

eirslett:
Maybe this package could replace the existing instrumentations for express/restify?

ghermeto:
my opinion is that using the underlying `http.IncomingMessage` and `http.ServerResponse` will be more stable and more likely to work with other frameworks (although it needs `koa-connect` module to work with Koa and won't work with Fastify).

@jcchavezs I really like your suggestions, so please wait for me to update the PR.

Regarding the deprecation, is there a timeline? Because I noticed today that the `zipkin-instrumentation-restify` was not working with Restify 7.2.3 due to a change on how middlewares are executed and I was going to submit a PR to fix it.

ghermeto:
@jcchavezs I looked at both your suggestions and here is what I found:

1) I will send a separate PR for that, because the best place to add the `defaultTags` is on the `instrumentation/httpServer.recordRequest` method. This way all the other server instrumentation can use it.

2) connect router (almost) literally [matches the lowered case string](https://github.com/senchalabs/connect/blob/master/index.js#L160) so for connect I can only do `GET /real/request/path`. For express, route matching [happens at runtime](https://github.com/expressjs/express/blob/master/lib/router/index.js#L136-L282), which means we won't have `req.route.path` available at the middleware. I could name the span `GET /user/123` instead, but if we already log the the `http.path` how valuable is that?

adriancole:
>
>
>    1.
>
>     route matching happens at runtime
>    <https://github.com/expressjs/express/blob/master/lib/router/index.js#L136-L282>,
>    which means we won't have req.route.path available at the middleware.
>    I could name the span GET /user/123 instead, but if we already log the
>    the http.path how valuable is that?
>
> yeah we certainly don't want to use normal path as span name as it would
imply high cardinality names. Let's only use route when indeed we know it
is (often we default to just the method when we can't get a route pattern)

>

ghermeto:
PR sent for the default tags: https://github.com/openzipkin/zipkin-js/pull/302

@adriancole agree. Should we table that until we find a better solution? It will probably implicate on a break of contract...

adriancole:
I'm in favor of solving distinct problems sure!

jcchavezs:
@ghermeto totally, as @adriancole said we should only use route based naming when we can access to it and avoid high cardinality names. For example, in `hapi` we can access to it just before sending the response. In Symfony (PHP), it is hard to resolve the route path so we prepare a cached version to do the trick (but it is not straighforward). I believe the best effort is good for me :).

ghermeto:
@jcchavezs can we merge this one as it is?

jcchavezs:
Sure we can, I am not in my laptop now so @adriancole could you please
merge it?

Den tor. 29. nov. 2018, 09:54 skrev Guilherme Hermeto <
notifications@github.com:

PR #302, opened 2018-11-29
recoding default tags on server instrumentation

ghermeto:
**What**: this is a simple change without contract break to allow the server instrumentation to save default tags at the creation time and record the tags with the request.

**Why**: this gives flexibility for users to set default tags they want recorded.

**The catch**: A few considerations:

1) I decided to send the tags at the constructor, because the `default tags` will already be available at the middleware bootstrap and decided to record them after the `http.path` tag for consistency.

2) why `for..in` instead of using a Map (and for..of) or iterate over the object keys? Well, using the following benchmark:

```javascript
var Benchmark = require('benchmark');
var util = require('util');
var suite = new Benchmark.Suite;


function record(key, val) {
    const x = {};
    x[key] = val;
}

suite.
add('forEach keys', function() {
    const obj = {
        foo: 'bar',
        banana: 'fruit',
        buick: 'car'
    };

    Object.keys(obj).forEach(tag => {
        record(tag, obj[tag]);
    });
}).
add('for..in', function() {
    const obj = {
        foo: 'bar',
        banana: 'fruit',
        buick: 'car'
    };

    for (const tag in obj) {
        if(obj.hasOwnProperty(tag)) {
            record(tag, obj[tag]);
        }
    }
}).
add('Map for..of', function() {
    const map = new Map();
    map.set('foo', 'bar');
    map.set('banana', 'fuit');
    map.set('buick', 'car');

    for (const [tag, value] of map) {
        record(tag, value);
    }
}).
on('cycle', function(event) {
    console.log(String(event.target));
}).
on('complete', function() {
    console.log('Fastest is ' + this.filter('fastest').map('name'));
}).
run();
```

I could verify that the `for..in` approach is 30% faster than the others. And because we must hit this middleware for all routes, performance counts.

cc/ @jcchavezs @adriancole

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -306,7 +306,13 @@ declare namespace zipkin {
 
   namespace Instrumentation {
     class HttpServer {
-      constructor(args: { tracer: Tracer, port: number });
+      constructor(args: {
+        tracer: Tracer,
+        port: number,
+        serviceName?: string,
+        host?: string,
+        serverTags?: {[key: string]: string}
+      });
 
       recordRequest(
         method: string,
--- a/packages/zipkin/src/instrumentation/httpClient.js
+++ b/packages/zipkin/src/instrumentation/httpClient.js
@@ -10,9 +10,11 @@ class HttpClientInstrumentation {
   constructor({
     tracer = requiredArg('tracer'),
     serviceName = tracer.localEndpoint.serviceName,
-    remoteServiceName
+    remoteServiceName,
+    clientTags = {}
   }) {
     this.tracer = tracer;
+    this.clientTags = clientTags;
     this.serviceName = serviceName;
     this.remoteServiceName = remoteServiceName;
   }
@@ -25,6 +27,11 @@ class HttpClientInstrumentation {
     this.tracer.recordServiceName(this.serviceName);
     this.tracer.recordRpc(method.toUpperCase());
     this.tracer.recordBinary('http.path', path);
+
+    if (this.clientTags) {
+      this.tracer.setTags(this.clientTags);
+    }
+
     this.tracer.recordAnnotation(new Annotation.ClientSend());
     if (this.remoteServiceName) {
       // TODO: can we get the host and port of the http connection?
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -84,11 +84,8 @@ class HttpServerInstrumentation {
     this.tracer.recordRpc(method.toUpperCase());
     this.tracer.recordBinary('http.path', path);
 
-    // eslint-disable-next-line no-restricted-syntax
-    for (const tag in this.serverTags) {
-      if (this.serverTags.hasOwnProperty(tag)) {
-        this.tracer.recordBinary(tag, this.serverTags[tag]);
-      }
+    if (this.serverTags) {
+      this.tracer.setTags(this.serverTags);
     }
 
     this.tracer.recordAnnotation(new Annotation.ServerRecv());
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -31,11 +31,13 @@ class HttpServerInstrumentation {
     serviceName = tracer.localEndpoint.serviceName,
     host,
     port = requiredArg('port'),
+    serverTags = {}
   }) {
     this.tracer = tracer;
     this.serviceName = serviceName;
     this.host = host && new InetAddress(host);
     this.port = port;
+    this.serverTags = serverTags;
   }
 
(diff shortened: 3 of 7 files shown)


adriancole:
main question is why here? for example, we have default tags requested
at tracer level quite a bit. What sort of example values would you be
using here that wouldn't be on the tracer? Ex some have asked for
site-specific tags and they wouldn't necessarily be http specific.
OTOH, I suppose middleware version related stuff could make sense.

Can you mention which tags you'd use this for?

Maybe we can use this advice for others (though in brave you can do
this anyway with HttpServerParser)

ghermeto:
@adriancole this is to address this comment from @jcchavezs: https://github.com/openzipkin/zipkin-js/pull/300#issuecomment-442378891 from the connect middleware PR. As far as I could understand, the `zipkin-go` implementation [provides something like this](https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L27).

I choose the server instrumentation instead of implementing it on the middleware because then other middleware, like `Hapi` would benefit from it too.

For Netflix specifically, we need to be able to add specific tags for our environment. Sure we can call the binary annotation, but it doesn't seem that unusual need.

adriancole:
Gotcha.. main thing I'm wondering is if this wouldn't be on all spans.
If so, it might be better placed in the tracer like this surprisingly
incomplete feature https://github.com/openzipkin/brave/issues/357

jcchavezs:
I think `defaultTags` or `serverTags` is something that should be addressed
per use case, e.g. server might have default tags different from SQL
instrumentation. Setting them at tracer level might be cumbersome.

Den tor. 29. nov. 2018, 23:01 skrev Adrian Cole <notifications@github.com:

adriancole:
>
> I think `defaultTags` or `serverTags` is something that should be addressed
> per use case, e.g. server might have default tags different from SQL
> instrumentation. Setting them at tracer level might be cumbersome.
>

right, though in the case of environments, it is kindof silly to copy them
to every subsystem. Main thing we'd want to document is how to customize
data for a subsystem. If here we want to have a separate function for not
request specific tags vs request specific tags, that can certainly make
sense. However, I would challenge us to use specific examples, as
environment constants such as application identifiers or node names will
not change subsystem to subsystem.. make sense?

>

adriancole:
one example feature we added was "finished span handler" which would
allow you to add tags or anything else given a mutable span, and a
trace context. With that, you can decide if something should only be
added conditional to another tag, if it was sampled, etc. There's no
infrastructure in place for this in JS at the moment, but anyway it is
food for thought.

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/FinishedSpanHandler.java

ghermeto:
sorry, I got lost... are there any actions for me to take? Should I move the code to the `tracer` or will we have it on the `http server instrumentation`?

adriancole:
primarily my question was to you as you were asking for the feature, @ghermeto  :)

is your use case solved by adding default tags to the tracer? If so, I think this is better as it isn't dependent on making changes here and later grpc, or any other middleware.

jcchavezs:
Beautiful @ghermeto. Shall we merge this? cc @adriancole

packages/zipkin/index.d.ts
```diff
@@ -306,7 +306,13 @@ declare namespace zipkin {
 
   namespace Instrumentation {
     class HttpServer {
-      constructor(args: { tracer: Tracer, port: number });
+      constructor(args: {
+        tracer: Tracer,
+        port: number,
+        serviceName?: string,
+        host?: string,
+        serverTags?: {[key: string]: string}
```
adriancole: do we want to qualify only server or just default? ex what about message consumer?

jcchavezs:
This is on http server middlewares, I guess the tags for consumer goes in
consumer.

Den man. 3. des. 2018, 19:20 skrev Adrian Cole <notifications@github.com:

packages/zipkin/index.d.ts
jcchavezs: I would keep it as serverTags because these are only tags for server.

ghermeto:
done!

jcchavezs:
Great @ghermeto

packages/zipkin/index.d.ts
adriancole: my intention was to hint you towards Tracer.defaultTags vs sporadically adding to every type of instrumentation.

The rationale is that this is default, there's no input parameter.. the more apis we add, the more we have to carry forward. I don't believe that a parameterless setting like this will serve us better than Tracer.defaultTags.

adriancole: so literally I was thinking you pass the tracer, you copy or take a reference to its default tags. then you get a single place to add and manage tags from. if there was a parameter type (ex request object etc), this wouldn't be sensible, but considering there is no scope here, and no example of tag that needs to know if it is a server or not, the simplest solution seems to be Tracer.defaultTags imho

PR #301, opened 2018-11-28
update libraries and scripts

ghermeto:
Reasoning: `zipkin-js` dev dependencies are far behind. 
This commit mainly updates `mocha` and `lerna`.

PR #303, opened 2018-11-29
adding gRPC client instrumentation

ghermeto:
**What**: It creates a `grpc` node client interceptor, as [proposed here](https://github.com/grpc/proposal/blob/master/L5-node-client-interceptors.md) and in line with the current [API documentation](https://grpc.io/grpc/node/module-src_client_interceptors.html).

It keeps consistency with the other instrumentation packages in `zipkin-js`, where the tracing must be provided in order to create the interceptor. It will be up to the client generator to abstract it from the end user.

@adriancole please review the annotations and tags being recorded. Am I recording the same as in other languages?

@drobertduke please review it too. I based this work on the interceptors you sent me, but I might be missing something.

cc/ @jcchavezs

### DIFF
--- a/packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
+++ b/packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
@@ -1,0 +1,116 @@
+const {Annotation, HttpHeaders} = require('zipkin');
+
+/**
+ * @private
+ * @param {string} name
+ * @throws Error
+ */
+function requiredArg(name) {
+  throw new Error(`GrpcClientInstrumentation: Missing required argument ${name}.`);
+}
+
+/**
+ * @namespace grpc
+ */
+
+/**
+ * @typedef {Object} grpc.Status
+ * @memberof grpc
+ * @property {function()} clone
+ * @property {function(key: string, value: string)} add
+ */
+
+/**
+ * @typedef {Object} grpc.Metadata
+ * @memberof grpc
+ * @property {function()} clone
+ * @property {function(key: string, value: string)} add
+ */
+
+/**
+ * @typedef {Object} GrpcClientContext
+ * @property {zipkin.Tracer} tracer
+ * @property {string} serviceName
+ * @property {string} remoteServiceName
+ */
+
+/**
+ * @class GrpcClientInstrumentation
+ */
+class GrpcClientInstrumentation {
+  /**
+   * @constructor
+   * @param {Object} grpc
+   * @param {GrpcClientContext} context
+   */
+  constructor(grpc, {
+    tracer = requiredArg('tracer'),
+    serviceName = tracer.localEndpoint.serviceName,
+    remoteServiceName
+  }) {
+    this.tracer = tracer;
+    this.serviceName = serviceName;
+    this.remoteServiceName = remoteServiceName;
+  }
+
+  /**
+   * Appends zipking headers to gRPC metadata
+   * @static
+   * @param {grpc.Metadata} originalMetadata
+   * @param {zipkin.TraceId} traceId
+   * @return {grpc.Metadata}
+   */
+  static setHeaders(originalMetadata, traceId) {
+    const metadata = originalMetadata.clone();
+    metadata.add(HttpHeaders.TraceId, traceId.traceId);
+    metadata.add(HttpHeaders.SpanId, traceId.spanId);
+
+    traceId._parentId.ifPresent(psid => {
+      metadata.add(HttpHeaders.ParentSpanId, psid);
+    });
+    traceId.sampled.ifPresent(sampled => {
+      metadata.add(HttpHeaders.Sampled, sampled ? '1' : '0');
+    });
+
+    return metadata;
+  }
+
+  /**
+   * Records start of RPC request
+   * @param {grpc.Metadata} metadata
+   * @param {string} method
+   * @return {grpc.Metadata}
+   */
+  start(metadata, method) {
+    this.tracer.setId(this.tracer.createChildId());
+    const traceId = this.tracer.id;
+
+    this.tracer.recordServiceName(this.serviceName);
+    this.tracer.recordRpc(method);
+    this.tracer.recordAnnotation(new Annotation.ClientSend());
+    if (this.remoteServiceName) {
+      this.tracer.recordAnnotation(new Annotation.ServerAddr({
+        serviceName: this.remoteServiceName
+      }));
+    }
+
+    return GrpcClientInstrumentation.setHeaders(metadata, traceId);
+  }
+
+  /**
+   * Records end of RPC request
+   * @param {zipkin.TraceId} traceId
+   * @param {grpc.Status} status
+   */
+  onReceiveStatus(traceId, status) {
+    const code = status.code.toString();
+    this.tracer.setId(traceId);
+    this.tracer.recordBinary('grpc.status_code', code);
+    if (code !== '0') {
+      this.tracer.recordBinary('error', code);
+    }
+    this.tracer.recordAnnotation(new Annotation.ClientRecv());
+  }
+}
+
+module.exports = GrpcClientInstrumentation;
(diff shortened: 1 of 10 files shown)


adriancole:
@ghermeto this is another work in progress. I think both you and @anuraaga would like to have grpc stuff merged. Maybe have a co-look at eachothers' stuff and we try to get 'em both merged? https://github.com/openzipkin/zipkin-go/pull/95

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+class GrpcClientInstrumentation {
+  /**
+   * @constructor
+   * @param {Object} grpc
+   * @param {GrpcClientContext} context
+   */
+  constructor(grpc, {
+    tracer = requiredArg('tracer'),
+    serviceName = tracer.localEndpoint.serviceName,
```
adriancole: it isn't great to set this specific to a component as it encourages people to have inconsistent names. better to inherit it for now, from the tracer, so that it doesn't end up mixed.

remoteServiceName for client code is legit

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
```diff
@@ -0,0 +1,116 @@
...
+   */
+  start(metadata, method) {
+    this.tracer.setId(this.tracer.createChildId());
+    const traceId = this.tracer.id;
+
+    this.tracer.recordServiceName(this.serviceName);
+    this.tracer.recordRpc(method);
+    this.tracer.recordAnnotation(new Annotation.ClientSend());
+    if (this.remoteServiceName) {
+      this.tracer.recordAnnotation(new Annotation.ServerAddr({
```
adriancole: possible you could get the intended socket IP here as well?

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Records end of RPC request
+   * @param {zipkin.TraceId} traceId
+   * @param {grpc.Status} status
+   */
+  onReceiveStatus(traceId, status) {
+    const code = status.code.toString();
+    this.tracer.setId(traceId);
+    this.tracer.recordBinary('grpc.status_code', code);
```
adriancole: looks like brave iirc

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
```diff
@@ -0,0 +1,116 @@
...
+    serviceName = tracer.localEndpoint.serviceName,
+    remoteServiceName
+  }) {
+    this.tracer = tracer;
+    this.serviceName = serviceName;
+    this.remoteServiceName = remoteServiceName;
+  }
+
+  /**
+   * Appends zipking headers to gRPC metadata
```
anuraaga: Typo: zipkin

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
anuraaga: Brave only tags when status is not OK.

https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java

Personally I would tag grpc.status_code always, like here and propose to do that in Brave too since the HTTP status has less meaning in grpc. What do you think?

jcchavezs:
Worth to involve @basvanbeek as he is a grpc guru.

Den tor. 29. nov. 2018, 19:24 skrev José Carlos Chávez <jcchavezs@gmail.com:

adriancole:
> - I would decouple the metadata propagation in GrpcClientInstrumentation
> from B3 because instead of passing a set of headers in the future we might
> want to implement the b3 single (double check @adriancole)
yep, though this can be a different change. Also, you'll notice in
brave we also try to read grpc-trace-bin!

> - Should we be able to just propagate the sampling decision since now
> sampling flags are a thing in zipkin-js? (If traceID and spanID are empty,
> again double check @adriancole). This is useful if you have a proxy making
> the sampling decision.
If we chose IDs we should probably propagate them as it is useful for
log correlation or when alternative sampling is going on. Otherwise,
you won't get consistent IDs in logs. Not everyone uses log
correlation, but quite a lot do.

> Great work @ghermeto <https://github.com/ghermeto>!
agree!

adriancole:
>
> Brave only tags when status is not OK.
>
>
> https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java
>
> Personally I would tag grpc.status_code always, like here and propose to
> do that in Brave too since the HTTP status has less meaning in grpc. What
> do you think?
>
tagging on success puts index pressure that's always the reason we have for
not adding tags. In brave it is less an issue as you can control the
parser. For example, there's an X-Ray one that chooses a whole bunch of
tags as index pressure is "not your problem" in SaaS :P

>

ghermeto:
> tagging on success puts index pressure that's always the reason we have for not adding tags. In brave it is less an issue as you can control the parser. For example, there's an X-Ray one that chooses a whole bunch of tags as index pressure is "not your problem" in SaaS :P

@adriancole currently the http server instrumentation in `zipikn-js` [tags the status code on success](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/instrumentation/httpServer.js#L92). Should that be changed?

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
ghermeto: unfortunately no... I checked with @drobertduke and that happens on a lower level. I can't access it from the interceptors.

adriancole:
> @adriancole currently the http server instrumentation in zipikn-js tags the status code on success. Should that be changed?
Personally, I think so.
https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L194

packages/zipkin-instrumentation-grpc-client/src/grpcClientInstrumentation.js
adriancole: thanks for checking

adriancole:
PS if there's a general interest in always tagging grpc status, I think it could be fine. We don't document it yet anyway! Plus our tagging policy is already very conservative vs opencensus for example, so will be quite efficient. All that said, we should keep in mind that data policy apis are a more flexible answer to what's otherwise a tug of war with multiple correct answers.

Regardless, let's just not reverse that back to an http default as it would imply a load increase for folks (under the assumption less use grpc than http). Not without a wider discussion anyway. https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L194

ghermeto:
### Updates:

1) decoupled setHeaders from start
2) removed serverName from params
3) fixed typo on comment
4) added `X-B3-Flags` header set to 1 **if traceId.isDebug()**
5) only tag with `grpc.status_code` **if status !== `grpc.status.OK`**

@adriancole I didn't do anything regarding `grpc-tracing-bin` yet because they are still experimental.
Also, I decided to only log `grpc.status_code` on error, because later it is easier to add it than to remove it...

adriancole:
SGTM thanks

ghermeto:
can we merge https://github.com/openzipkin/zipkin-js/pull/301 before this one? I need to update the `yarn.lock` here and they will both conflict.

anuraaga:
Not tagging on success SGTM, thanks for clarifying

ghermeto:
can anyone re-run the tests? Seems like we have a flaky test on the request instrumentation...

adriancole:
kicked

ghermeto:
it passed! 👍

adriancole:
lovely! now we can try 3 different grpc instrumentation together, once the zipkin-go one is in!

PR #304, opened 2018-11-30
send x-b3-flags header with http if debug is true

ghermeto:
**Why**: on the [discussion about gRPC client](https://github.com/openzipkin/zipkin-js/pull/303#issuecomment-442940269), @jcchavezs brought to my attention I should be sending `X-B3-Flags` if debug is true. I later noticed http were not sending that header either.


adriancole:
thanks much

PR #306, opened 2018-12-03
Porting Zipkin package to TypeScript

DavideCarvalho:
Since #290 is too far behind from the master, I created a new branch so we can start fresh.
We had a problem on #290 that when building the lib, the file extensions weren't been deleted, so if we required a ts file on ```/src```, it will be required when built as well.
Doing some tests, looks like Import and Export statements do this for us, so I think we can do this in order to port zipkin package.

### DIFF
--- a/packages/zipkin/.babelrc
+++ b/packages/zipkin/.babelrc
@@ -1,3 +1,6 @@
 {
-  "extends": "../../.babelrc"
+  "extends": "../../.babelrc",
+  "presets": [
+    "@babel/preset-typescript"
+  ]
 }
--- a/packages/zipkin/package.json
+++ b/packages/zipkin/package.json
@@ -5,8 +5,8 @@
   "main": "lib/index.js",
   "types": "index.d.ts",
   "scripts": {
-    "build": "babel src -d lib",
-    "test": "tsc index.d.ts && mocha --require ../../test/helper.js",
+    "build": "babel src -d lib --extensions \".ts\" --extensions \".js\"",
+    "test": "tsc index.d.ts && mocha --require ../../test/helper.js --require @babel/register",
     "prepublish": "npm run build"
   },
   "author": "OpenZipkin <openzipkin.alt@gmail.com>",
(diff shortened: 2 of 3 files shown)


jcchavezs:
Great Idea. Would be nice to have input from @openzipkin/zipkin-js-champions

sonybinhle:
Yes, I also think that changing from `require('options.js')` to `import option from 'options.ts'` is the best way to solve the problem. Typescript only understands and can transpile `import` statement properly. 

There will not be any breaking changes assuming that all other packages import files from node_modules:

1. We only transpile `src` code to `lib` code like before, the only difference is using `babel-typescript` preset.
2. We didn't change public APIs, we only change `require` to `import`
3. If all tests passed.

webmutation:
This could not be better timming, need to import zipkinjs to angular app, the typing will help. Any issues converting the current browserify setup to webpack... is anyone working on this?

My goal is by the end of next week to have a simple angular cli app using zipkinjs.

DavideCarvalho:
We have a issue to use Rollup to bundle zipkinjs, if you want, give it a try.

Enviado do meu iPhone

Em 4 de dez de 2018, à(s) 20:07, DDd <notifications@github.com> escreveu:

webmutation:
Thanks for your reply David.

Will take a look tomorrow, I am more familiar with parcel and webpack than rollup, but can take a look... 

I changed the example web browser.js to use import instead of require and using parcel it works fine... just playground for work tomorrow https://github.com/webmutation/zipkin-js-example/tree/master/web 

had to upgrade  'browser-process-hrtime'  to 1.0.0 since typings are not in previous version.

Will take a look at the other js files tomorrow to convert them to ES2016 module then fully to TS with your work.

jcchavezs:
Awesome work @DavideCarvalho!

jcchavezs:
@webmutation this is the issue we were talking about: https://github.com/openzipkin/zipkin-js/issues/270

DavideCarvalho:
@jcchavezs I've migrated some more code to ts, so instead of doing a big bang merge request, I'll creating a bunch of merges so we do not lose track and keep rebasing every single day the typescript branch, are you okay with it?

PR #308, opened 2018-12-05
Fixing imports of ts files and tests

DavideCarvalho:
I've migrated 3 files to ```.ts``` and fixed the tests, and I think people should be careful with one thing: Instead of using ```require``` to import ```.ts``` files, use ```import``` instead.


DavideCarvalho:
@sonybinhle could you help me out with it? ```zipkin-instrumentation-cujojs-rest``` tests are breaking on the ci, but locally they are passing.

webmutation:
Hi Davi, you probably found it already the logs are there seems like typing problems
https://travis-ci.org/openzipkin/zipkin-js/jobs/463833959

sonybinhle:
@DavideCarvalho 
We may change `import * as Tracer from './tracer';` to `import Tracer from './tracer';` because we export only one Tracer class from `tracer/index.js`. 

If we import Tracer using the wildcard(*) syntax, Tracer will be an object { default: function } and not be a function which can be instantiated by `new` keyword.

DavideCarvalho:
@webmutation those typing errors do not make the test fail, as far as I know. The test is on zipkin-instrumentation-cujojs-rest, but I can't reproduce it locally
@sonybinhle I've changed already.

webmutation:
It looks to me it fails linting first. Don't know settings are used in travis. But failing a linting the build is failed.

"The command "yarn lint" exited with 1."

Then in another stage also fails the "zipkin-instrumentation-cujojs-rest".

"The command "yarn test" exited with 1."

So two issues to be fixed.

jcchavezs:
Ping @eirslett as you were the original author of that package.

DavideCarvalho:
I’ve tested on two computers locally and none of them gave me an error, even after @sonybinhle change.

jcchavezs:
I just tried in my computer and it failed:

image

My node version is v9.11.1 tho.

PS. Before I run this, I installed everything with yarn install.

jcchavezs:
@DavideCarvalho so [last run](https://travis-ci.org/openzipkin/zipkin-js/jobs/463942544) did not fail because of the tests but the lint:

image

DavideCarvalho:
Okay, so the tests are running, so we need to investigate what’s going on with the ci.

Lint will always output error, since we are migrating the code, but this shouldn’t break the test.

jcchavezs:
Well, this is regular behaviour from travis I think, see this post:
https://medium.com/@manjula.cse/how-to-stop-the-execution-of-travis-pipeline-if-script-exits-with-an-error-f0e5a43206bf

If this is going to be a regular problem we could disable temporary the
lint but I would prefer that every PR like this someone else could open a
PR fixing the lint.

Shall we just merge this?

Den man. 10. des. 2018, 11:18 skrev Davi de Carvalho <
notifications@github.com:

DavideCarvalho:
Let's merge tomorrow, I'd like to try a little more today

packages/zipkin/src/annotation.ts
```diff
@@ -14,78 +14,80 @@ class ProducerStop extends SimpleAnnotation {}
...
+function BinaryAnnotation(key, value): void {
   this.key = key;
   this.value = value;
 }
-BinaryAnnotation.prototype.toString = function() {
+BinaryAnnotation.prototype.toString = function (): string {
   return `BinaryAnnotation(${this.key}="${this.value}")`;
 };
 
+new ServiceName('teste');
```
eirslett: I guess this one should be removed?

packages/zipkin-instrumentation-cujojs-rest/test/integrationTest.js
```diff
@@ -1,4 +1,5 @@
-const {Tracer, ExplicitContext} = require('zipkin');
+// const {Tracer, ExplicitContext} = require('zipkin');
+import { Tracer, ExplicitContext } from 'zipkin';
 const express = require('express');
```
eirslett: Could we just change all the requires in the whole project to import, all at once? (I think there are codemods for it)

jcchavezs:
@DavideCarvalho any chance we can relaunch this?

jcchavezs:

PR #310, opened 2018-12-07
#270: Add rollup for packages/zipkin to build es, cjs, umd

sonybinhle:
The PR is created for testing the changes in CI and asking for early reviews.

- [x] Propose rollup build for packages/zipkin

- [x] CI build success for packages/zipkin


DavideCarvalho:
I tthink those 3 formats are all we need.
Rollup tree-shaking is experimental, and I don't think we would beneficiate the applications with it, so don't bother.

jcchavezs:
I am about to merge this by tomorrow.

Cc @eirslett

Den lør. 8. des. 2018, 02:03 skrev Davi de Carvalho <
notifications@github.com:

jcchavezs:
@sonybinhle I have a question, is this going to impact the releasing process or we can just merge it as it is?

DavideCarvalho:
I think rollup script should use some kind of minifier too, this will make the bundle smaller

sonybinhle:
> I think rollup script should use some kind of minifier too, this will make the bundle smaller
From what I observed in some popular library how they used Rollup.

1. They have two versions of UMD(development and production) and the production version is minified.
2. They don't minify es6, cjs because either it will be used in Node or it will be used through some bundler like Webpack, Parcel(minification step is here).

The two statements above are not true with some library like React where production bundle has significant less code then minified one. (They eliminated much development only code)

I think in our case, maybe minification for cjs format is unnecessary?

jcchavezs:
I think it is worth to include @ewhauser on this discussion.

sonybinhle:
> @sonybinhle I have a question, is this going to impact the releasing process or we can just merge it as it is?

@jcchavezs I think it will not affect the releasing process because the cjs built is still under `lib/index.js` and the `.npmignore` file doesn't contain `dist` or `es` folder.

jcchavezs:
So to be clear, who will be in charge of generating the `dist` files on every release?

jcchavezs:
Maybe I did not ellaborate my question in the best way:

The current flow is:
1. Merge code into master
2. run yarn publish and it gets to npm directory

So if we are bundling files, I believe there should be an step where we run
`yarn build` or `yarn prepublish`. If that is true, where it fits? Could we
automatize? If not, just ignore me :p.

Ping @openzipkin/zipkin-js-champions @ghermeto

José Carlos Chávez


Den man. 10. des. 2018 kl. 10:25 skrev Binh Le <notifications@github.com>:

jcchavezs:
Reason why I ask about this is because I would prefer to either have an
automated think running the build on release time (not sure we will be able
to test it) or as a commit before merge. My main concern is this becoming
manual and we have to ask people to do so every time we merge something or
before we release something. Does that make sense?

Den man. 10. des. 2018, 10:38 skrev José Carlos Chávez <jcchavezs@gmail.com:

DavideCarvalho:
we can set prepublish script to build, and build to generate the bundle, this way everytime we publish it will make the bundle automatically.

jcchavezs:
Sounds good. Could we address that here @sonybinhle?

Den man. 10. des. 2018, 11:14 skrev Davi de Carvalho <
notifications@github.com:

sonybinhle:
@jcchavezs It is already added to our current package.json. `prepublish": "npm run build"`
If we run `yarn publish`, it will automatically run `prepublish` first.

jcchavezs:
I see. Thanks for the clarification.

Let's wait for @ewhauser feedback about the browser concerns.

ewhauser:
+1 on the change; I’m going to be doing a bunch of integration testing on
this stuff over the next few weeks so can submit fixes if any issues are
found

jcchavezs:
Great, then we will merge this by tomorrow.

Den man. 10. des. 2018, 17:34 skrev Eric Hauser <notifications@github.com:

jcchavezs:
@sonybinhle could you please rebase?

sonybinhle:
@jcchavezs It is rebased.

jcchavezs:
Thanks a lot @sonybinhle!

José Carlos Chávez


Den tir. 11. des. 2018 kl. 16:19 skrev Binh Le <notifications@github.com>:

Issue #311, opened 2018-12-08
Make browser support a first class citizen


ewhauser:
I'm integrating a browser app and I'm finding a couple of issues related to browser support. I'd like to open a number of sub-issues related to making browser support work again, but wanted feedback prior to doing it:

1. Update the README to explicitly state which modules are supported by under which runtimes
2. For browser supported modules, all tests should be run under karma as well as mocha to prevent browser regressions
3. Fix `BatchRecorder` to not have a hard dependency on `node-fetch`
4. Fix `InetAddress` to not have a hard dependency on `os`

It appears 3 & 4 *may* be fixed in head but they are broken as of 0.14.3 which is what is on npm. 1 & 2 will flesh that out.

@openzipkin/zipkin-js-champions

jcchavezs:
We are totally up for this.

1 & 2 are things we definitively want,
3 could be fixed but in my experience in other languages it is alright if
we have to use two libraries for http reporting. So if at some point we
need this to be configurable it is OK.
4. Is something @openzipkin/zipkin-js-champions and @ghermeto
<https://github.com/ghermeto>could provide feedback about.

Den lør. 8. des. 2018, 22:17 skrev Eric Hauser <notifications@github.com:

jcchavezs:
> This is great, but we have some stuff ahead in order to release version
1, I don’t know if we have time for it right now. If the community helps,
it will make stuff a lot easier

+1 to this, we depend on community work.

Den søn. 9. des. 2018, 01:22 skrev Eric Hauser <notifications@github.com:

Issue #54, opened 2016-10-12
We should probably update the README to talk about browsers


adriancole:
The README focuses on node, but we do actually work on browsers. Maybe we could revise some things and suggest browserify (if that's a decent recommendation).

adriancole:
suppose we should also link the example to the README, too https://github.com/openzipkin/zipkin-js-example

jcchavezs:
Closed by https://github.com/openzipkin/zipkin-js/pull/312

PR #312, opened 2018-12-08
update readme to be more specific about browser vs nodejs support #54

ewhauser:
This updates the `README` file to be more explicit about which parts of this project are available for node.js and the browser. This addresses #54 and is the first task in #311.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,15 +48,19 @@ const zipkinFetch = wrapFetch(fetch, {tracer, remoteServiceName});
 
 The `zipkin` library can be used in the browser. The `web` [example](https://github.com/openzipkin/zipkin-js-example) shows an example of a browser based application making a call to a backend server with trace headers attached.
 
-## Instrumentation
+### Instrumentation
 
 The following libraries can be instrumented in the browser:
 
 - [fetch](packages/zipkin-instrumentation-fetch) (zipkin-instrumentation-fetch)
 
-## Transports
+### Transports
 
- Currently, the only way to record spans in the browser is to use the `ConsoleRecorder`: 
+The following transport is available for use in the browser:
+
+- [http](packages/zipkin-transport-http)
+
+For debugging purposes, you can also use the `ConsoleRecorder`:
 
 ```javascript
 const tracer = new Tracer({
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,7 +44,30 @@ const remoteServiceName = 'youtube';
 const zipkinFetch = wrapFetch(fetch, {tracer, remoteServiceName});
 ```
 
-## Instrumentations
+## Browser
+
+The `zipkin` library can be used in the browser. There are a few [examples](https://github.com/openzipkin/zipkin-js-example) available for the `web` and `react-native`.
```
adriancole: should we mention react native?

jcchavezs:
I would also ask you to write an example in
https://github.com/openzipkin/zipkin-js-example.

The idea on this is not only that it works an examples for uses but here we
also run E2E tests against the latest zipkin version every night to make
sure we don't break examples.

Den lør. 8. des. 2018, 23:37 skrev Eric Hauser <notifications@github.com:

ewhauser:
Not sure what you are asking for here. There already is an example in the project for the browser?

README.md
ewhauser: I was assuming the react-native example worked in the browser as well, but I see that it is an Android and iOS example only. I’ll remove this.

ewhauser:
Got it. That makes sense.

Once #314 is merged, all the unit tests will be running under both Firefox and Chrome which will provide integration tests for running under the browser.

As far as I can tell, there currently isn't a way to send local traces generated in the browser directly to a Zipkin server as the HTTP transport relies on node libraries. I'm planning on opening a separate issue to address that problem and think it would make sense to add an end to end test as part of that work.

ewhauser:
Did a bit more digging and it seems that the first revision was inaccurate. It does appear that you can use the `HttpLogger` in the browser. The issue I was running into was when compiling it with TypeScript for an Angular application. The file correctly has the logic to use `window.fetch` instead of `node-fetch` at runtime, but the AOT compiler was trying to load the `require` statements.

I found a workaround for this and will update.

jcchavezs:
@ewhauser shall we merge this?

ewhauser:
Yes, good to go

webmutation:
@ewhauser so you have an Angular application working with zipkin-js ? i am currently working on this, but if this has been done I will stop to prevent duplication of efforts.

jcchavezs:
Great!

José Carlos Chávez


Den tir. 11. des. 2018 kl. 16:11 skrev DDd <notifications@github.com>:

ewhauser:
@webmutation - Yes, I have a prototype that integrates Zipkin and Angular. I'm finishing up the README and tests but plan on making it available by the end of the week.

webmutation:
Oh great news. Can you share the repo as is if its public.

ewhauser:
@webmutation I've made this available here: https://github.com/ewhauser/angular-tracing. Please keep in mind that it is an early prototype and there are probably dragons. Feel free to provide some feedback!

Issue #313, opened 2018-12-08
Run tests under karma to prevent browser regressions


ewhauser:
The tests currently only run via a node.js runner which allows for the introduction of regressions when running under the browser.

jcchavezs:
Shall we close this as #314 is merged?

PR #314, opened 2018-12-08
run tests for zipkin library under karma #313

ewhauser:
Fix for #313. The setup works as follows:

- If `TEST_SUITE` env var is set to `nodejs`, then the mocha tests run
- If `TEST_SUITE` env var is set to `browser` then the karma tests run
- If `TEST_SUITE`env var is not set, then both suites run

A 3rd entry is added to the test matrix for `browser` in the CI runs which only runs the browser tests.

Note that there is a single test that was previously failing under the browser (unrelated to this change). I'm opening a separate issue to address this, but do not believe this should be a blocker in merging this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ matrix:
       node_js: 8
       env:
         - CXX=clang-3.6
-        - KAFKA_PLEASE_LOG=verbose
         - TEST_SUITE=browser
       addons:
         chrome: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+dist: trusty
+sudo: required
 
 matrix:
   include:
@@ -6,11 +8,22 @@ matrix:
       env:
         - CXX=clang-3.6
         - KAFKA_PLEASE_LOG=verbose
+        - TEST_SUITE=nodejs
 
     - node_js: 8
       env:
         - CXX=clang-3.6
         - KAFKA_PLEASE_LOG=verbose
+        - TEST_SUITE=nodejs
+
+    - name: 'browser'
+      node_js: 8
+      env:
+        - CXX=clang-3.6
+        - KAFKA_PLEASE_LOG=verbose
+        - TEST_SUITE=browser
+      addons:
+        chrome: stable
 
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.27.5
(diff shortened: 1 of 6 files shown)


jcchavezs:
I think this is all right. @davidecarvalho is making a big effort on
porting the whole code to TS so in case we have some failures or need
test/input when something fails in karma I would like to get some input
from you too. Is that OK with you?


Den søn. 9. des. 2018, 05:29 skrev Adrian Cole <notifications@github.com:

jcchavezs:
@ghermeto I will merge this and include it in 0.15. Do you see any
challenge?

Den søn. 9. des. 2018, 07:40 skrev José Carlos Chávez <jcchavezs@gmail.com:

jcchavezs:
Actually there is no rush to merge this, I would rather wait for feedback
from @davidecarvalho, @sonybinhle and @ghermeto before and this could be
potentially part or 0.16 (if that is a thing) so we can see first how this
PR behaves with @davidecarvalho changes.

Den søn. 9. des. 2018, 20:07 skrev José Carlos Chávez <jcchavezs@gmail.com:

ewhauser:
In #312, I've clarified that a couple of additional libraries can also be run under the browser: `zipkin-instrumentation-fetch` and `zipkin-transport-http`. The tests for these libraries are written in a different way (as integration tests requiring the spawning of a node.js server), so refactoring those to work under `karma` will take a bit more work. Assuming this gets merged, I'll open separate issues for dealing with those dependencies.

.travis.yml
```diff
@@ -1,16 +1,29 @@
...
       env:
         - CXX=clang-3.6
         - KAFKA_PLEASE_LOG=verbose
+        - TEST_SUITE=nodejs
+
+    - name: 'browser'
+      node_js: 8
+      env:
+        - CXX=clang-3.6
+        - KAFKA_PLEASE_LOG=verbose
```
ghermeto: I thought you weren't running node tests... why the kafka log setup?

ewhauser:
@ghermeto Please see my note above. Adding `karma` tests for the additional libraries is more difficult due to the fact that they are written as integration tests operating under the assumption that they are running in a nodejs context. My preference would be to get this merged and then refactor to running all the tests.

ghermeto:
@ewhauser makes sense to split in multiple commits

👍

adriancole:
thanks a lot @ewhauser! and @ghermeto for reviewing

Issue #316, opened 2018-12-10
Autopublish merges into master as canaries


jcchavezs:
Currently, releasing is a manual process, not often and error prone. It is very likely changes in configuration can break the process without further notice. Releasing every canary would fix that problem.

See https://github.com/lerna/lerna/tree/master/commands/publish#--canary

Ping @openzipkin/zipkin-js-champions

DavideCarvalho:
We should not use master canary releases. Master should only publish when we have a pack of changes ready to be publish as a stable version.

Instead of using Master for that, we should create a “develop” branch and use it for that. Why develop? See gitflow https://zellwk.com/blog/git-flow/

eirslett:
I disagree, I don't think having a separate develop branch and a "release-only" master branch is a good idea. Git tags serve the purpose of identifying which commits are release commits. I think every commit on the master branch should be released as a canary. So :+1:

ghermeto:
would be much appreciated. 👍

adriancole:
sounds fine to me as long as versions do sensible things. master should always be working though there is some big change with typescript coming up so common sense should apply on when to merge things knowing the canary stuff. (ex increment version before huge change)

devinsba:
I agree with the canaries from master idea, if only because it aligns this repo with the rest of the org, in that releases are done on tag and master is the most up to date snapshot

DavideCarvalho:
I don't agree. Master should have the latest stable version. Usually, when using gitflow, canary releases are released on develop branch.

eirslett:
i think the Zipkin core server is released as a snapshot/canary on every commit/merge to master, is that still the case @adriancole ? That worked pretty well.

It's true that with gitflow uses a develop branch and a separate master branch: I'm against using gitflow at all, because we don't need the value it provides in exchange for extra complexity. I prefer canary releases from master, and stable releases on git tags. (Only my 2c though)

jcchavezs:
Thanks everyone for the feedback. My thoughts on this:

I think adding canaries for release provide many benefits:
1. Exercise the release process.
2. Publish non stable tags so people could try stuff and shorten the
feedback loop before we provide a proper release.

That does not mean we encourage people to use this canary releases (if
someone wants to use it is at its own risk).

In the other hand, using different branches is something inconvenient from
my POV:
1. It causes that changes are integrated later and that cause massive
rebases (think that zipkin-js is not something people work on every day, as
a volunteering project, people work on it when they have time). The sooner
[...]

PR #317, opened 2018-12-10
fix: fixes lerna script for releasing.

jcchavezs:
With the upgrade of lerna, this script for releasing got outdated. There is an [issue](https://github.com/openzipkin/zipkin-js/issues/316) to prevent this in the future.

Ping @ghermeto.


adriancole:
👍

PR #319, opened 2018-12-13
Fix express middleware - fixes #318

### LINKED ISSUE #318: Should TraceId be set to express req.headers or res.locals if it's empty at the beginning?
According to issue https://github.com/openzipkin/zipkin-js/issues/88, express middleware doesn't seem to be able to work well to capture following middleware's operation's span especially if it involves async/await/Promise work.

When in expressMiddleware, it starts `instrumentation.recordRequest` and generated one id, would it be more helpful for the developers to deal with different scenarios if it's set into the req.headers or res.locals so that they can pick up the context in later middleware in order to be enclosed in the same trace as child span?

ghermeto:
**Why?** the tracer scoped inside the middleware was preventing the traceId to be accessed from sequential promises in the route handler.

@jcchavezs I added a test that verifies that the tracerId will remains the same within the same request, but will be different on subsequent requests.

### DIFF
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -1,9 +1,11 @@
-const {
-  option: {Some, None},
-  Instrumentation
-} = require('zipkin');
+const {option: {Some, None}, Instrumentation} = require('zipkin');
 const url = require('url');
 
+/**
+ * @private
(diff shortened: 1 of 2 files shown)


jcchavezs:
Awesome!

landons:
Is there a process for tagging NPM releases? This was merged over a month ago, and I'd love to use the _trace_id getter when this is available.

PR #320, opened 2018-12-13
adds route.path to express instrumentation

ghermeto:
adds route.path to rpc when possible. In express if the request
is terminated in the middleware (like in express-http-proxy)
the request won't have the route property set.

cc/ @adriancole @jcchavezs 

this replaces #200


ghermeto:
argh! I need to fix this flaky test some day.. what a pain...

PR #322, opened 2019-01-03
Do not send redis command payloads

voldern:
The tracing of single commands do not record the payload but batched commands include it.

For many commands with big payloads this can end up becoming several megabytes of annotation information.

This commit removes it for batched commands as well.


jcchavezs:
Thanks @voldern. Ping @adriancole & @pbadenski

adriancole:
good catch. agreed we are not a redis storage option :)

jcchavezs:
Tusen takk @voldern!

PR #325, opened 2019-01-11
Upgraded the node-fetch version to 2.3.0 in zipkin-transport-http

Kiho:
Upgraded the node-fetch version to 2.3.0 in zipkin-transport-http.
We are getting build warning  from react application because of this old dependency.

### DIFF
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -13,7 +13,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "node-fetch": "^1.5.3"
+    "node-fetch": "^2.3.0"
   },
   "devDependencies": {
     "@babel/cli": "7.1.5",


jcchavezs:
Thanks @Kiho let's keep this until tomorrow and see if someone from @openzipkin/zipkin-js-champions or @ghermeto give some feedback.

ghermeto:
simple enough, although, in the long tun, [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch) makes more sense to better support using this transport module from the browser.

Looks good. 👍

Kiho:
I found another solution here for our problem https://github.com/webpack/webpack/issues/3078#issuecomment-400697407
, but still I like to have solution from zipkin side.

Thanks.

jcchavezs:
> simple enough, although, in the long tun, [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch) makes more sense to better support using this transport module from the browser.

Could you please create an issue for that?

PR #328, opened 2019-02-12
httpServer instrumentation should let sampler to make decision if missing sampled header

### LINKED ISSUE #327: Should httpServer instrumentation assumes sampled header always provided with spanId?
In HttpServerInstrumentation, the logic of creating traceId when header contains x-b3-spanid and x-b3-traceid assumes that the x-b3-sampled headers are also provided.

https://github.com/openzipkin/zipkin-js/blob/9e0eee22a07156056656c2eeca8b4f94d5dbed34/packages/zipkin/src/instrumentation/httpServer.js#L47

the value of `sampled` is actually `None` when we are using the `readHeader` function provided from `zipkin-instrumentation-connect` middleware.

Then it results in that the traceId created by `zipkin-instrumentation-connect` middleware will always not captured.

Should the `None` value be directly used here?  Would it be better if it has similar logic when calling `tracer.createRootId` without parameters, which calls the `sampler` to decide?

The issue raised here is that the Pivotal Cloud Foundry which has Zipkin enabled only injected those three ids header without the sampled header.

https://docs.pivotal.io/pivotalcf/2-4/concepts/http-routing.html#http-headers

kenspirit:
Closes #327

### DIFF
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -54,6 +54,11 @@ class HttpServerInstrumentation {
           flags
         });
       });
+
+      const parentIdValue = parentId.getOrElse();
+      if (parentIdValue._sampled === None) {
+        parentIdValue._sampled = this.tracer.sampler.shouldSample(parentIdValue);
+      }
       return !this.tracer.supportsJoin
         ? parentId.map(id =>
             this.tracer.letId(id, () =>
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -55,17 +55,7 @@ class HttpServerInstrumentation {
         });
       });
 
-      const parentIdValue = parentId.getOrElse();
-      if (parentIdValue._sampled === None) {
-        parentIdValue._sampled = this.tracer.sampler.shouldSample(parentIdValue);
-      }
-      return !this.tracer.supportsJoin
-        ? parentId.map(id =>
-            this.tracer.letId(id, () =>
-              this.tracer.createChildId()
-            )
-          )
-        : parentId;
+      return this.tracer.join(parentId);
     } else {
       if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
         const sampled = readHeader(Header.Sampled) === None ?
--- a/packages/zipkin/src/instrumentation/httpServer.js
+++ b/packages/zipkin/src/instrumentation/httpServer.js
@@ -55,7 +55,7 @@ class HttpServerInstrumentation {
         });
       });
 
-      return this.tracer.join(parentId);
+      return new Some(this.tracer.join(parentId.getOrElse()));
     } else {
       if (readHeader(Header.Flags) !== None || readHeader(Header.Sampled) !== None) {
         const sampled = readHeader(Header.Sampled) === None ?
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -157,7 +157,7 @@ class Tracer {
 
   join(traceId) {
     if (!(traceId instanceof TraceId)) {
-      return this.createRootId();
+      throw new Error('Must be valid TraceId instance');
     }
 
     if (traceId._sampled === None) {
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -156,21 +156,19 @@ class Tracer {
   }
 
   join(traceId) {
-    if (traceId === None) {
-      return new Some(this.createRootId());
+    if (!(traceId instanceof TraceId)) {
+      return this.createRootId();
     }
 
-    const value = traceId.getOrElse();
-    if (value._sampled === None) {
-      value._sampled = this.sampler.shouldSample(value);
+    if (traceId._sampled === None) {
+      /* eslint-disable no-param-reassign */
+      traceId._sampled = this.sampler.shouldSample(traceId);
     }
 
     return !this.supportsJoin
-      ? traceId.map(id =>
-        this.letId(id, () =>
+      ? this.letId(traceId, () =>
           this.createChildId()
         )
-      )
       : traceId;
   }
 
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -157,7 +157,7 @@ class Tracer {
 
   join(traceId) {
     if (traceId === None) {
-      return this.createRootId();
+      return new Some(this.createRootId());
     }
 
     const value = traceId.getOrElse();
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -155,6 +155,25 @@ class Tracer {
     });
   }
 
+  join(traceId) {
+    if (traceId === None) {
+      return this.createRootId();
+    }
+
+    const value = traceId.getOrElse();
+    if (value._sampled === None) {
+      value._sampled = this.sampler.shouldSample(value);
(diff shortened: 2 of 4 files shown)


jcchavezs:
I like this much better, could we add tests for tracer too?

kenspirit:
> I like this much better, could we add tests for tracer too?

Currently the scenario of backfilling and sampled value calculation is in  httpServerInstrumentation.test.js, should I move those into tracer or add new ones?

jcchavezs:
@kenspirit yes please, since we need to test that unit.

packages/zipkin/src/tracer/index.js
```diff
@@ -155,6 +155,25 @@ class Tracer {
     });
   }
 
+  join(traceId) {
+    if (traceId === None) {
+      return new Some(this.createRootId());
```
adriancole: why would we return an option type here? would it not be easier to return a trace ID not wrapped?

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -54,13 +54,8 @@ class HttpServerInstrumentation {
...
       });
-      return !this.tracer.supportsJoin
-        ? parentId.map(id =>
-            this.tracer.letId(id, () =>
-              this.tracer.createChildId()
-            )
-          )
-        : parentId;
+
+      return this.tracer.join(parentId);
```
adriancole: you'll notice all the below return a Some() and this is an internal function. Probably, unless I missed something, we can return just a trace ID here to simplify

packages/zipkin/src/tracer/index.js
kenspirit: This is just to align to the same value type return below.  Hence, I try to reply both to your comment here.

The `join` method provided in tracer is intended to backfilling the span/trace created from upstream.  The `readHeader` method in HttpServer instrumentation return Some/None wrapping the spanId/traceId.  Hence, the input of traceId should normally be Some/None.

Then for below condition branch, possible original pass in traceId is returned or a childId wrapped in Some is returned.

Hence, to align input datatype or return value, I use Some to wrap it.

To be honest, I am not sure if there is any guideline on when Some/None should be used to wrap the raw value.  Please let me know if there is any and my comment above does not handle reasonably.

adriancole:
if there is some tech debt folks dont feel comfortable about changing in
http internal methods that is ok to punt. they are internal anyway.

however we shouldn't add tech debt to new methods on tracer. joinSpan
should not have a condition which results in a None result.

kenspirit:
Sorry that I don't quite understand your opinion.  Please enlighten me a bit.

>however we shouldn't add tech debt to new methods on tracer. joinSpan should not have a condition which results in a None result.

The `join` method doesn't want to results in None result and that is why I put the first checking there.

`tracer.join` should be mainly used for backfilling external service's traceId (such as from http headers), right?  If the traceId is local service, we should use createChildId method for it.

Then for those external service's traceId, would we normally expect them to be raw traceId or wrapped in option type which used Some/None to express instead of TraceId/null?

If we expect the input to be option type, should the return value with same format giving that in some condition we are providing its original value?

adriancole:
I think I understand what you are asking.. that if there is a scenario
where join simply returns its input unaffected.. and if it did should we
not let the caller know (via option val of Some I guess)

I would respond two ways.

1. if there was a case like that, it might be temporary. For example, in
brave we add the flag shared when we are sharing an ID.
2. even if we did return the same value, the code wouldnt likely behave any
differently

imho option parameters on top level functions or option results are
confusing as they beg the question why. so rather that ask why we shouldn't
be returning option, I tend to ask why should we.

does above clarify at all?

kenspirit:
My major point is not on returning unaffected value but whether option type value is some differentiation of TraceId for local/external service.  I originally thought this kind of wrapping should be used for that.

However, I got your concern that top level function using option type is confusing after I further check other top level functions in other module.

adriancole:
not to derail, but fyi

the span kind is used to handle knowledge of something being remote.

there are some different actionable signals we use in brave related as
well: localRootId (this tells you we are at a sub graph) and shared (this
tells you we are sharing an ID.

while only the "shared" flag applies here  (which doesnt yet exist in
zipkin-js iirc) the other things above might be of general interest to you.

glad you are interested!

kenspirit:
definitely, thanks for sharing.

packages/zipkin/test/trace.test.js
```diff
@@ -417,11 +417,11 @@ describe('Tracer', () => {
     expect(rootTracerId.traceId.length).to.eql(32);
   });
 
-  it('should create new TraceId when joining None', () => {
+  it('should create new TraceId when joining none TraceId value', () => {
     const {recorder, ctxImpl} = setupTest();
     const tracer = new Tracer({recorder, ctxImpl});
 
-    const newTraceId = tracer.join(None).getOrElse();
+    const newTraceId = tracer.join(null);
```
adriancole: I don't think we should allow join on null should we?

kenspirit: Yes, we don't.  So that is why in such case, it creates a new root id.  Should we assumes null will never pass in?  This test case just to ensure that join method works for all none TraceId value, especially null.

adriancole:
Yes, we don't. So that is why in such case, it creates a new root id.
> Should we assumes null will never pass in? This test case just to ensure
> that join method works for all none TraceId value, especially null.


code that passes in null would be buggy code. I would throw an error if you
want to cover this with code.

jcchavezs:
I am happy with this. Let's keep it open until tomorrow to see if we get further feedback from @openzipkin/zipkin-js-champions

jcchavezs:
@kenspirit could you please rebase so I can merge it?

kenspirit:
@jcchavezs reset local and rebase to do a new commit with force push to my fork and issue a new PR?  I am not so familiar this process.  Last time I tried this, the PR got closed due to my force push.

If you don't mind, please kindly share the detail step I should do.  I roughly remember then PR merge option should have "Rebase and merge"?

jcchavezs:
Thanks @kenspirit

kenspirit:
welcome.  My honor.

adriancole:
last commit made some breakage as callers cannot guarantee the object will match type TraceId due to transpilation See https://github.com/openzipkin/zipkin-js/issues/422

kenspirit:
> last commit made some breakage as callers cannot guarantee the object will match type TraceId due to transpilation See #422

OMG.  Do miss this case.  Looks like the `instanceof` should be used very cautious due to thhe transpilation is commonly used now.

adriancole:
I'm also a bit scared about optional type as it uses instanceof internally. I wonder if we can create a test that internally transpiles vs simulating? cc @abesto

abesto:
Huh. This whole situation is weird. I don't see how the the breakage is introduced. AFAICT there is no way for client code to have any other `TraceId` type from the one exposed by `zipkin-js`. So that must mean we expose multiple types, and the `instanceof` check is using one, while callers are using another. Weird.

Trying to track this down I started messing around with 00604b7 (the commit that was merged). Especially weird is this line in the generated `packages/zipkin/es/index.js`:

```
var TraceId_1 = TraceId;
```

I find the same in `dist/zipkin.js`. I do *not* find `TraceId_1` in `dist/zipkin.min.js`, which may be a smoking gun or a red herring.

> I wonder if we can create a test that internally transpiles vs simulating?

Agree something like this would be useful. Without it, I don't even have a good way of reproducing the issue. The cleanest way might be to run tests both as we do now _and_ after minification, but I couldn't find any Babel / Mocha docs on doing that.

Maybe some of you can easily reproduce the issue? As I was researching other folks hitting similar issues, I came across https://babeljs.io/docs/en/next/babel-plugin-transform-instanceof.html which seems like it might possibly "fix" this and let us use `instanceof` at will.

PR #333, opened 2019-02-15
should make a local span with defaultTags

alexgervais:
Expecting the `defaultTags` to be recorded on spans created by local trace.

As per the failing test case:
```
  1) Tracer
       should make a local span with defaultTags:

      AssertionError: expected { Object (key, value) } to deeply equal { Object (serviceName) }
      + expected - actual

       {
      -  "key": "instanceId"
      -  "value": "i-1234567890abcdef0"
      +  "serviceName": "smoothie-store"
       }
```
The default tag is recorded, but not as part of the `trace.local()` operation generating the span.

### DIFF
--- a/packages/zipkin/test/trace.test.js
+++ b/packages/zipkin/test/trace.test.js
@@ -141,6 +141,43 @@ describe('Tracer', () => {
     });
   });
 
+  it('should make a local span with defaultTags', () => {
+    const record = sinon.spy();
+    const recorder = {record};
+    const ctxImpl = new ExplicitContext();
+    const localServiceName = 'smoothie-store';
+    const defaultTags = {instanceId: 'i-1234567890abcdef0', cluster: 'nodeservice-stage'};
+    const trace = new Tracer({ctxImpl, recorder, localServiceName, defaultTags});
+
+    ctxImpl.scoped(() => {
+      const result = trace.local('buy-smoothie', () => {
+        trace.recordBinary('taste', 'banana');
+        return 'smoothie';
+      });
+
+      expect(result).to.eql('smoothie');
+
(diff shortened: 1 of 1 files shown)


jcchavezs:
Ping @ghermeto

packages/zipkin/test/trace.test.js
```diff
@@ -141,6 +141,43 @@ describe('Tracer', () => {
...
+
+    ctxImpl.scoped(() => {
+      const result = trace.local('buy-smoothie', () => {
+        trace.recordBinary('taste', 'banana');
+        return 'smoothie';
+      });
+
+      expect(result).to.eql('smoothie');
+
+      expect(record.getCall(0).args[0].annotation).to.eql(
```
ghermeto: I expect this test to always fail because the calls are made in a different order from this test. 

The correct order is: 
```
[ 
  BinaryAnnotation { key: 'instanceId', value: 'i-1234567890abcdef0' },
  BinaryAnnotation { key: 'cluster', value: 'nodeservice-stage' },
  ServiceName { serviceName: 'smoothie-store' },
  LocalOperationStart { name: 'buy-smoothie' },
  BinaryAnnotation { key: 'taste', value: 'banana' },
  LocalOperationStop {} 
]
```

The defaultTags comes first because [it is recorded in the constructor](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/tracer/index.js#L54-L56). Then,  `local` will [record the service name and the other annotations](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/tracer/index.js#L114-L115).

If you want tags to be recorded on a different order you can always call `tarce.setTags` instead.

cc/ @jcchavezs

jcchavezs:
Closed by #334

Issue #330, opened 2019-02-12
Tracer defaultTags behavior and documentation


alexgervais:
Using `zipkin-js` 0.16.0, I tried to understand how the Tracer `defaultTags` property by looking at the original PR and test cases.

Given the following code:
```
	const ctxImpl = new (require('zipkin').ExplicitContext)();
	const recorder = new (require('zipkin').BatchRecorder)({
		logger: new (require('zipkin-transport-http')).HttpLogger({
			endpoint: `${ZIPKIN_BASE_URL}/api/v2/spans`,
			jsonEncoder: require('zipkin').jsonEncoder.JSON_V2
		})
	});

	const tracer = new (require('zipkin').Tracer)({
		ctxImpl: ctxImpl,
		recorder: recorder,
		localServiceName: 'my-app',
		defaultTags: {
			version: '1.0.0'
		}
	});
```

Spans are sent and collected by OpenZipkin. However, they are all missing the expected `version: 1.0.0` tag.

Unable to find the related documentation for Tracer configuration, I'm not sure if my usage if right and this is a bug, or I'm simply not using the feature correctly.

jcchavezs:
Indeed spans should be tagged with default tags and they should be added on span creation. I guess the issue is in https://github.com/openzipkin/zipkin-js/pull/302/files#diff-ac3b9e6d0bce19fd854267694cd42421R108 because in the test we add the tag instead of expect the tracer to automatically add it. Am I wrong @ghermeto ?

ghermeto:
@jcchavezs AFAIK it works just fine:

https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/test/trace.test.js#L122-L142

jcchavezs:
@alexgervais would you raise a failing test case for this?

alexgervais:
I created a failing test case: https://github.com/openzipkin/zipkin-js/pull/333
Again, it might just be my understanding of the feature that's not in line with the implemented behavior.

ghermeto:
I added a comment on the PR. The constructor will record the `defaultTags` first.

@jcchavezs  I will add a test to find out why @alexgervais is unable to find the tags in `openZipkin`.

ghermeto:
Ok, definitely a **bug** @alexgervais.

The problem is precisely because `setTags` is being called before the `traceId` is set. In this case, `BatchRecorder` will update a different partial span.

[...]

PR #334, opened 2019-02-16
fix(Tracer): defaultTags were not being propagated

ghermeto:
**What?**
This fix an issue with the Tracer defaultTags param to make sure
the tags are in the correct span scope. Issue was discussed on #330 

**Why?**
defaultTags were being recorded on the Tracer constructor and that
is usually before the Tracer ID is set. In this case BatchRecorder
would record the tags in different PartialSpan (the one that has
the default id from the Tracer).

**How?**
* keeping the Tracer constructor API unaltered
* adds new method: recordDefaultTags
* calls recordDefaultTags on httpClient, httpServer and Tracer.local
* also adds convenience getter defaultTags

**Who?**
@jcchavezs @adriancole @alexgervais

### DIFF
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -8,6 +8,12 @@ const {Span, Endpoint} = require('./model');
 const defaultTimeout = 60 * 1000000;
 
 /**
+ * defaultTags property name
+ * @type {symbol}
+ */
+const defaultTagsSymbol = Symbol('defaultTags');
+
+/**
  * @class PartialSpan
  */
 class PartialSpan {
@@ -33,6 +39,26 @@ class PartialSpan {
     }
     this.endTimestamp = now(this.startTimestamp, this.startTick);
   }
+
+  /**
+   * factory: creates new span and set
+   * @static
+   * @param {TraceId} id
+   * @param {Object} defaultTags
+   * @return {PartialSpan}
+   */
+  static create(id, defaultTags = {}) {
+    const span = new PartialSpan(id);
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const tag in defaultTags) {
+      if (defaultTags.hasOwnProperty(tag)) {
+        span.delegate.putTag(tag, defaultTags[tag]);
+      }
+    }
+
+    return span;
+  }
 }
 
 /**
@@ -49,6 +75,7 @@ class BatchRecorder {
     this.logger = logger;
     this.timeout = timeout;
     this.partialSpans = new Map();
+    this[defaultTagsSymbol] = {};
 
     // read through the partials spans regularly
     // and collect any timed-out ones
@@ -90,7 +117,7 @@ class BatchRecorder {
     if (this.partialSpans.has(id)) {
       span = this.partialSpans.get(id);
     } else {
-      span = new PartialSpan(id);
+      span = PartialSpan.create(id, this[defaultTagsSymbol]);
     }
     updater(span);
     if (span.endTimestamp) {
@@ -182,6 +209,10 @@ class BatchRecorder {
           break;
       }
     });
+  }
+
+  setDefaultTags(tags) {
+    this[defaultTagsSymbol] = tags;
   }
 
   toString() {
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -1,25 +1,51 @@
 const {now, hrtime} = require('./time');
 const {Span, Endpoint} = require('./model');
 
-function PartialSpan(traceId) {
-  this.traceId = traceId;
-  this.startTimestamp = now();
-  this.startTick = hrtime();
-  this.delegate = new Span(traceId);
-  this.localEndpoint = new Endpoint({});
-}
-PartialSpan.prototype.finish = function finish() {
-  if (this.endTimestamp) {
-    return;
-  }
-  this.endTimestamp = now(this.startTimestamp, this.startTick);
-};
+/**
+ * default timeout = 60 seconds (in microseconds)
+ * @type {number}
+ */
+const defaultTimeout = 60 * 1000000;
 
+/**
+ * @class PartialSpan
+ */
+class PartialSpan {
+  /**
+   * @constructor
+   * @param {TraceId} traceId
+   */
+  constructor(traceId) {
+    this.traceId = traceId;
+    this.startTimestamp = now();
+    this.startTick = hrtime();
+    this.delegate = new Span(traceId);
+    this.localEndpoint = new Endpoint({});
+  }
+
+  /**
+   * adds endTimestamp to span
+   * @return {void}
+   */
+  finish() {
+    if (this.endTimestamp) {
+      return;
+    }
+    this.endTimestamp = now(this.startTimestamp, this.startTick);
+  }
+}
+
+/**
+ * @class BatchRecorder
+ */
 class BatchRecorder {
-  constructor({
-    logger,
-    timeout = 60 * 1000000 // default timeout = 60 seconds
-  }) {
+  /**
+   * @constructor
+   * @param {Object} options
+   * @property {HttpLogger|KafkaLogger} logger logs the data to openZipkin
+   * @property {number} timeout timeout for span in microseconds
+   */
+  constructor({logger, timeout = defaultTimeout}) {
     this.logger = logger;
     this.timeout = timeout;
     this.partialSpans = new Map();
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -68,7 +68,7 @@ class BatchRecorder {
   /**
    * @constructor
    * @param {Object} options
-   * @property {HttpLogger|KafkaLogger} logger logs the data to openZipkin
+   * @property {Logger} logger logs the data to openZipkin
    * @property {number} timeout timeout for span in microseconds
    */
   constructor({logger, timeout = defaultTimeout}) {
--- a/packages/zipkin/src/console-recorder.js
+++ b/packages/zipkin/src/console-recorder.js
@@ -1,10 +1,4 @@
 /**
- * defaultTags property name
- * @type {symbol}
- */
-const defaultTagsSymbol = Symbol('defaultTags');
-
-/**
  * @class ConsoleRecorder
  */
 class ConsoleRecorder {
@@ -15,21 +9,14 @@ class ConsoleRecorder {
 
   record(rec) {
     const {spanId, parentId, traceId} = rec.traceId;
-    const defaultTags = this[defaultTagsSymbol] || {};
-    const tags = JSON.parse(defaultTags);
-
     this.logger(
       `Record at (spanId=${spanId}, parentId=${parentId}, ` +
-      `traceId=${traceId}, defaultTags=${tags}): ${rec.annotation.toString()}`
+      `traceId=${traceId}): ${rec.annotation.toString()}`
     );
   }
 
   toString() {
     return 'consoleTracer';
-  }
-
-  setDefaultTags(tags) {
-    this[defaultTagsSymbol] = tags;
   }
 
 }
--- a/packages/zipkin/src/console-recorder.js
+++ b/packages/zipkin/src/console-recorder.js
@@ -1,19 +1,37 @@
+/**
+ * defaultTags property name
+ * @type {symbol}
+ */
+const defaultTagsSymbol = Symbol('defaultTags');
+
+/**
+ * @class ConsoleRecorder
+ */
 class ConsoleRecorder {
    /* eslint-disable no-console */
   constructor(logger = console.log) {
     this.logger = logger;
   }
+
   record(rec) {
-    const id = rec.traceId;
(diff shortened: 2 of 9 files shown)


adriancole:
I kindof perceive this as a temporary hack. is there a convention we can use to hide this method? I was not expecting users to ever call this ( in a way http abstraction is a user ), rather, the recorder object itself.

ghermeto:
well, if the `defaultTags` lives on the `Tracer`, then we just wouldn't have access to the `defaultTags` on the [recorder method](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L81-L159) of `BatchRecorder` in the context of the correct `TraceId`.

Either way, httpServer and httpClient need to be the one to trigger `setTags` (directly or indirectly), because they set the tracer ID.

another idea is to replace the calls to `recordServiceName` in the server and client code by a method like:

```js
  recordService(serviceName) {
    this.recordAnnotation(
      new Annotation.ServiceName(serviceName)
    );
    this.setTags(this.defaultTags);
  }
```

but I fail to see how this is any better...

Do you have any other suggestions @adriancole ?

adriancole:
I think doing this on service is better because it doesn't expose a new
function. Generally I am opposed to patchwork functions as people might
rely on them. all RPC (in fact all spans) should set service name.

adriancole:
if we ever make a v2 api this stuff will be easy to contain inside a span.

jcchavezs:
Hi @ghermeto thanks a lot for the great work. I have a couple of questions:

1. Do we need to expose `defaultTags` from the tracer given we have the `recordDefaultTags` method? https://github.com/openzipkin/zipkin-js/pull/334/files#diff-aa30be809cfe81ece1c873af6639ef3cR171
2. Since `recordDefaultTags` is not something that should be exposed in the API, could we make sure that is clear? Maybe calling it `__recordDefaultTags` and adding a comment in there could help.

ghermeto:
@jcchavezs @adriancole what about if I:

- [ ] remove the method `recordDefaultTags`
- [ ] create a `recordService` method in `Tracer`:
```js
  recordService(serviceName) {
    this.recordAnnotation(
      new Annotation.ServiceName(serviceName)
    );
    this.setTags(this.defaultTags);
  }
```
- [ ] replace all calls to `recordServiceName` to `recordService` in `httpServer`, `httpClient` and `Tracer.local`

### Other considerations:

#### Why keep the getter `defaultTags`? 

We want to make sure that user don't alter the `defaultTags` after the tracer is created, but we should allow them to read them. 

#### Why not just modify `recordServiceName`?

`recordServiceName` is a very specific name and that recording `defaultTags` is an unexpected side-effect. And simply renaming `recordServiceName` to `recordService` is a breaking change of the API.

adriancole:
I think we are investing too much time in an api that should have been
wholesale rewritten at least a year ago. we really need to start thinking
about redo and less patchwork. in other words, the tracer api is flawed in
general and same energy could be used towards making a new one.

why not one of my original suggestions which is to move it to the recorder
object? to be very specific, add a method to the BatchRecorder with default
tags. When a tracer is constructed with the batch recorder set this? In
hind sight we should have started with this.

adriancole:
another reason. service name should be constant for all spans. the fact
that it isnt and this api style affords arbitrary service naming results in
poor quality data and bad aggregation. adding another method (which implies
advertising it) will not likely reduce people's misuse.


the service name and default tags should be added consistently to all spans
iotw. the only thing that can do anything consistently is batch recorder.
the reasons is that the old circa 2012 tracer design was written like a
logger.. stateless except for scoping of a trace id. unless we are making
an intentionally hidden side effect in tracer, we might as well get it
closer to right and put it on the batch recorder until someone makes time
to do this ecosystem the favor of adding a span api.

my 2p

adriancole:
one other hack could be to add the state of default tags to the traceid
itself. then the batch recorder can inspect it, and add these when the span
is first used there.

this is a hack that will result in more memory use but light feel cleaner
to those not comfortable with the tracer passing initialization state to
its batch recorder.

I can't think of any other alternatives so I will leave this to you folks.
even if you decide to do something I dont like it is fine as you are the
people supporting and using it. I hope feedback was helpful tho as I have
endured many bad data caused by this style of tracer and we have a chance
to make bad data less frequent.

ghermeto:
@adriancole that is a very good point. This API needs a rewrite, but while 1.0 is not released I wouldn't start working on 2.0 just yet.

Ok, assuming we want to add defaultTags `BatchRecorder`, here is what I suggest:

1. Make a method on `BatchRecorder` called, for instance, `setDefaultTags`:
```javascript
  setDefaultTags(tags) {
    this[defaultTagsSymbol] = tags;
  }
```
2. calls that method on the `Tracer` constructor (have to add that method on `ConsoleRecorder` too)
```javascript
  this.recorder.setDefaultTags(defaultTags);
```
3. add a method `putTags` on `Span` that will take multiple tags (`or` we can add a private method on `BatchRecorder` that will call `Span.putTag` multiple times, although that doesn't looks like a good design decision).
```javascript
Span.prototype.putTags = function putTags(tags = {}) {
  // eslint-disable-next-line no-restricted-syntax
  for (const tag in tags) {
    if (tags.hasOwnProperty(tag)) {
      this.putTag(tag, tags[tag]);
    }
  }
};
```
4. on `BatchRecorder`, every time we call `Span.sertName` (namely `Rpc` and `LocalOperationStart`) we call the method to set the `defaultTags`
```javascript
         case 'LocalOperationStart':
          span.delegate.setName(rec.annotation.name);
          span.delegate.putTags(this[defaultTagsSymbol]);
          break;
        case 'LocalOperationStop':
          span.finish();
          break;
        case 'Message':
          span.delegate.addAnnotation(rec.timestamp, rec.annotation.message);
          break;
        case 'Rpc':
          span.delegate.setName(rec.annotation.name);
          span.delegate.putTags(this[defaultTagsSymbol]);
          break;
```

@jcchavezs do you agree with this approach?

adriancole:
Why wouldn't you just add the tags at the beginning (preferably) like here:
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L67

or at the end like here:
https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L15

on putTags, I don't see why we need to increase the surface area..

Are you concerned that doing putTag 3 times will be a performance concern?
Is that needed to accomplish the goal?

adriancole:
To be specific, the looping in tags should only need to occur once... when they are applied, which is likely best when you know you are creating a new span. With that in mind, separating out a private method which is only called once is my question.. It isn't a blocker or anything, but yeah.. usually I personally wait until code is reused before extracting things but anyway this is style and not a blocker comment.

jcchavezs:
So I am more going for not increasing the api surface otherwise we have more API to maintain in next versions. I agree the API need a refactor tho we are still learning about the user needs and issues in JS (and definitively @ghermeto inputs and help is priceless).

That said, I would choose a hack here with comments in code. My first vote would be for keeping `recordServiceName` and do the magic there but not expose defaultTags on tracer (I don't see the use case for user needing to read back them) or what @adriancole suggested:
> Why wouldn't you just add the tags at the beginning (preferably) like here: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L67 or at the end like here: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/batch-recorder.js#L15

What do you think @ghermeto would you please explain what do you think about @adriancole's proposal?

ghermeto:
I think @adriancole suggestion of setting the tags when the span is created is a good idea. I will use that. 

One thing though, since setting the default tags will be part of the span creation, I think it makes sense to create a static method on `PartialSpan` to handle it. It doesn't **"increase surface"** because `PartialSpan` is not exposed. And I don't think we can use the fact that this code needs refactoring to double down on its issues. If anything, we need to modernize the code as much as we can, make it simpler, so that the refactor is easier.

I can remove the getter for defaultTags too. If that becomes necessary we will know.

Now here is my question: AFAIK recorder has no interface, but we have two recorders (`BatchRecorder` and `ConsoleRecorder`), but no parent class. Now that I need to add `setDefaultTags` to the recorder I find myself in need to add the method in two classes. Can I create a `Recorder` abstract class or make `BatchRecorder` inherit from `ConsoleRecorder`?

ghermeto:
I will just make that part of a different PR and we can discuss this further.

adriancole:
> Now here is my question: AFAIK recorder has no interface, but we have two
> recorders (BatchRecorder and ConsoleRecorder), but no parent class. Now
> that I need to add setDefaultTags to the recorder I find myself in need
> to add the method in two classes. Can I create a Recorder abstract class
> or make BatchRecorder inherit from ConsoleRecorder?
>
 since ConsoleRecorder is stateless, I'm not sure what you'd be able to do
there. How about just checking if the type is BatchRecorder? Or did you
have an idea of how to implement the same in ConsoleRecorder?

packages/zipkin/src/batch-recorder.js
```diff
@@ -1,28 +1,81 @@
...
+   * @static
+   * @param {TraceId} id
+   * @param {Object} defaultTags
+   * @return {PartialSpan}
+   */
+  static create(id, defaultTags = {}) {
+    const span = new PartialSpan(id);
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const tag in defaultTags) {
```
adriancole: 👍

packages/zipkin/src/console-recorder.js
```diff
@@ -1,19 +1,37 @@
...
-    const id = rec.traceId;
+    const {spanId, parentId, traceId} = rec.traceId;
+    const defaultTags = this[defaultTagsSymbol] || {};
+    const tags = JSON.parse(defaultTags);
+
     this.logger(
-      `Record at (spanId=${id.spanId}, parentId=${id.parentId},` +
-      ` traceId=${id.traceId}): ${rec.annotation.toString()}`
+      `Record at (spanId=${spanId}, parentId=${parentId}, ` +
+      `traceId=${traceId}, defaultTags=${tags}): ${rec.annotation.toString()}`
```
adriancole: I think this will dramatically increase logging volume, depending on default tags. I'd advise against this.

packages/zipkin/src/tracer/index.js
```diff
@@ -51,9 +51,7 @@ class Tracer {
     this._defaultTraceId = this.createRootId();
     this._startTimestamp = now();
     this._startTick = hrtime();
-    if (defaultTags) {
-      this.setTags(defaultTags);
-    }
+    this.recorder.setDefaultTags(defaultTags);
```
adriancole: maybe check for the method `setDefaultTags` or check the type is `BatchRecorder` for now?

packages/zipkin/src/console-recorder.js
ghermeto: what is the real world use for `ConsoleRecorder`?

adriancole:
people use things in surprising ways. some sites do scrape logs. I think it
is unnecessary risk to change the ConsoleRecorder  less to discuss also ;)

ghermeto:
thanks for the feedback @adriancole!

packages/zipkin/src/batch-recorder.js
```diff
@@ -1,28 +1,81 @@
...
+ */
 class BatchRecorder {
-  constructor({
-    logger,
-    timeout = 60 * 1000000 // default timeout = 60 seconds
-  }) {
+  /**
+   * @constructor
+   * @param {Object} options
+   * @property {HttpLogger|KafkaLogger} logger logs the data to openZipkin
```
jcchavezs: Minor nit, could we avoid this implementation detail `HttpLogger|KafkaLogger` unless it is absolutely necessary?

packages/zipkin/src/batch-recorder.js
```diff
@@ -64,7 +117,7 @@ class BatchRecorder {
     if (this.partialSpans.has(id)) {
       span = this.partialSpans.get(id);
     } else {
-      span = new PartialSpan(id);
+      span = PartialSpan.create(id, this[defaultTagsSymbol]);
```
jcchavezs: <3

packages/zipkin/src/batch-recorder.js
ghermeto: well, I would love to have a real common interface for `HttpLogger|KafkaLogger` but we don't right now. I will fix with a `@typedef`.

jcchavezs:
Could we have an interface?

José Carlos Chávez


Den tor. 21. feb. 2019 kl. 10:34 skrev Guilherme Hermeto <
notifications@github.com>:

packages/zipkin/src/batch-recorder.js
ghermeto: @jcchavezs if I replace by this, would it work?

```js
/**
 * @typedef {Object} ZipkinLogger
 * @property {function(Span):void} logSpan
 */

```

then use as:

```js
/**
 * @property {ZipkinLogger} logger logs the data to openZipkin
 */
```

ghermeto: scratch that... I think we have a `ts` interface already...

jcchavezs:
Thanks @ghermeto !

Issue #209, opened 2018-05-10
Collision of traceid when heavily loaded


jemapi:
Hi,

We're experiencing collisions of traceId on production when our microservices are heavily loaded, we assume that it's because the traceId is two base64 numbers concatenated, and those are generated with a non-collision-safe algorithm (like uuid).

Will be great that instead of support 64 / 28 bits traces (spans are allways 64), we can gave the implementation (function) to generate the traceId and spanId, this can give more flexibility to the lib. 

If the above is too much complex or doesn't append to the lib logic, then will be great that the traceId / spanId algorithm will be implemented collision-safe to avoid this issues.

Also, UUid is more expensive that what u guys are using, so in terms of performance it should impact.

Let me know your thoughts about it,

Greetings,

jcchavezs:
+1 to avoid collisions with a better algorithm. @jemapi would you come up with a PR exploring a solution?

jcchavezs:
I am closing this issue until we can prove collisions with a test (could be stress test). Feel free to reopen this in such case. Thanks @rcoedo for bringing this up.

adriancole:
Notably, if there is a problem in `randomTraceId()` we should fix it and benchmark it. externalizing isn't a fix. Regardless, we should be very careful about conclusions as in the last few weeks I noticed many scope leak bugs which could appear same as trace collision.

PR #336, opened 2019-02-21
allow id generators to be passed to the tracer (#209)

rcoedo:
This PR tries to address the issue #209.

The tracer now defines two extra optional function parameters, the traceId generator function and the spanId generator function. The spanId generator parameter defaults to the `randomTraceId` function that we were already using. The traceId generator parameter defaults to a new `randomTraceId128bit` function that mimics the current behaviour: `randomTraceId() + randomTraceId()`.

I am not sure this is the best approach but it is the best solution I could find without breaking the API or changing the current behaviour. I lack the deep understanding of zipkin to feel comfortable making deeper changes, so any suggestions will be welcomed 😄

cheers!

### DIFF
--- a/packages/zipkin/src/index.ts
+++ b/packages/zipkin/src/index.ts
@@ -3,7 +3,7 @@ export { default as option } from './option';
 export { default as Annotation } from './annotation';
 export { default as Tracer } from './tracer';
 export { default as createNoopTracer } from './tracer/noop';
-export { default as randomTraceId } from './tracer/randomTraceId';
+export { randomTraceId, randomTraceId128bit } from './tracer/randomTraceId';
 export { default as sampler } from './tracer/sampler';
 export { default as TraceId } from './tracer/TraceId';
 
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -4,7 +4,7 @@ const {Sampler, alwaysSample} = require('./sampler');
 const Annotation = require('../annotation');
 const Record = require('./record');
 const TraceId = require('./TraceId');
-const randomTraceId = require('./randomTraceId');
+const {randomTraceId, randomTraceId128bit} = require('./randomTraceId');
 const {now, hrtime} = require('../time');
 const {Endpoint} = require('../model');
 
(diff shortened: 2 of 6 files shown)


jcchavezs:
This looks legit to me, tho I would like to see how you did improve the traceID randomness and maybe add it as the standard way to create IDs? Ping @adriancole @ghermeto

rcoedo:
I think the randomness would improved by using something like `uuid` to generate the ids, but it would probably affect performance. With the current approach a developer would be able to choose their own algorithm if needed.

If you think the current default id generation is not good enough, I could default to `uuids` instead of the `randomTraceId` function.

jcchavezs:
I see. I though you tweaked the existing randomTraceId. It makes sense we
make it pluggable. I think UUID is great but for not too high load you
might not need that much randomness.

José Carlos Chávez


Den tor. 21. feb. 2019 kl. 10:19 skrev Roman Coedo <notifications@github.com

adriancole:
my opinion is that we should vet that random isnt working properly and maybe cite why. the issue leading to this sounded a bit like a hypothesis and we probably shouldn't add new APIs based on that rationale.

sometimes people think id collision is happening when really it is resource leak (like a TLS thing)

it is possible but unlikely that the actual problem is that two 64 bit numbers are clashing in a short amount of time.

anyway tldr is please consider not changing public api or making more knobs until the problem is vetted and something like this . I would also suggest rule of 3. like one other high volume site like netflix

adriancole:
sorry meant CLS not TLS (like id leaking on thread which is a frequent root cause when folks start pointing fingers at RNG because it has the same effect)

adriancole:
ps I was mentioned and so did a "drive by" suggestion. I'm feeling under the weather so very likely not to continue this thread in the next day.. iotw dont block on me  just consider and decide how to proceed!

ghermeto:
@adriancole I will ask Insights if they have instrumentation to check for collisions, but I personally haven't seen it so far.

jcchavezs:
After a second thought I also believe it is very unlike that we have collisions in generated traces and it looks to me more what Adrian says about CLS. I will close this issue unless @rcoedo is terribly sure this is the case (maybe with a high load test).

Issue #157, opened 2017-11-14
Instrumentation for "got"


voldern:
I've created an instrumentation for [got](https://www.npmjs.com/package/got) that can be found here: https://github.com/voldern/zipkin-instrumentation-got.

If enough people are interested and the maintainers think it's a good idea I can move it here.

FYI @sindresorhus

jcchavezs:
@voldern @sindresorhus there is this on-going work for got: https://github.com/openzipkin/zipkin-js/pull/341 thanks to @juhovh

juhovh:
Just for clarification, my version is using hooks that are only available in the latest version, instead of having to rewrite the API in the wrapper as done in the older version above. So in that sense it's much less invasive, but does require a recent version.

Issue #338, opened 2019-02-21
Drop support for node 6


jcchavezs:
Support for node 6 [is going to end on April 2019](https://nodejs.org/en/about/releases/). We should consider drop support for that after that happens.

There is an ongoing work on this matter: https://github.com/openzipkin/zipkin-js/pull/146

Ping @openzipkin/zipkin-js-champions

jcchavezs:
Ping @ghermeto

jcchavezs:
Duplicated https://github.com/openzipkin/zipkin-js/pull/146

jcchavezs:
Related to https://github.com/openzipkin/zipkin-js/pull/250

ghermeto:
👍 on it

eirslett:
👍 This will also let us default to using async hooks instead of continuation-local-storage.

jcchavezs:
So what are the actions we could take on this one? I think the first step is to remove node 6 from travis (I can do that). Could you @ghermeto and @eirslett create (or give visibility) to those actions that we could start?

PR #341, opened 2019-03-02
Add instrumentation for the Got HTTP client for Node.js.

juhovh:
[Got](https://github.com/sindresorhus/got) is a lightweight and composable HTTP client for Node.js that supports extending the client through hooks. This PR adds zipkin-js instrumentation to Got similar to the existing fetch API instrumentation.

To explain some implementation details, Got is fully asynchronous and supports for multiple retries for a single call. However, it does have an `init` hook called once for each call that is guaranteed to be synchronous. This implementation stores the caller (parent) trace context from Tracer during the synchronous init call and uses that context for each request separately to make sure the retries are siblings to each other (as they should be).

The tests are basically copied from the fetch instrumentation with some Got specific modifications.

Closes https://github.com/openzipkin/zipkin-js/issues/157


juhovh:
Hmm, it turns out we might have a problem here... The Got library has required Node.js version 8 since Got version 9.x, but I cannot easily downgrade to Got version 8.x because it does not support the `got.extend` API necessary for using the hooks. Please let me know what's the best way to resolve this, I can also release this as a separate library if integrating with zipkin-js is not a reasonable option...

jcchavezs:
Hi @juhovh,

Thanks for working on this. There is an issue (
https://github.com/openzipkin/zipkin-js/issues/338) for dropping support
for node 6 so I hope we do this in the near future.

For the moment, could you disable the unit tests for this package in node 6?

As feedback to this PR I would recommend to add the ts definitions to be
consistent with other packages.

Den lør. 2. mar. 2019, 06:01 skrev juhovh <notifications@github.com>:

juhovh:
Thank you @jcchavezs for a quick reply again. I had to add a `--ignore-engines` to the `yarn install` in Travis, because otherwise yarn would just fail in the engine version check coming from the got library.

I also ended up defining got as any in the ts definitions, it seems that even `@types/got` is a bit out of date and the next version of got is supposed to be written in TypeScript so the problem should eventually fix itself.

I also had to make the got require conditional because the library uses object spreads that are not supported in older versions of Node. Decided to just skip the tests if the library is not available, although I noticed that many existing tests are just using `process.exit(0)` in case the tests should be skipped. I just liked having the skipped tests in the test runner output even if the Node.js version is too small.

juhovh:
Oh and forgot to mention, I also added a `peerDependencies` check for got `>= 9.6.0` because I realized that the latest version 9.6.0 is the only version supporting the synchronous `init` hook necessary to get the trace context reliably from the global tracer object.

jcchavezs:
I am not comfortable disabling things in CI, what if instead push for https://github.com/openzipkin/zipkin-js/issues/338? Ping @ghermeto

jcchavezs:
Hi @juhovh I think your original PR should work now as [we dropped node6](https://github.com/openzipkin/zipkin-js/pull/344) from the supported engines.

juhovh:
@jcchavezs thanks for pinging me, I reverted the Node.js 6 specific changes but left some other changes (peer dependency, type definitions).

jcchavezs:
Are changes in `yarn.lock` necessary?

jcchavezs:
LGTM, what about you @ghermeto?

juhovh:
Oh, missed the question about `yarn.lock` changes being necessary, I guess they are never really necessary, but just let me know if any changes are needed.

jcchavezs:
@juhovh did you check about the yarn.lock changes? I would like to merge this PR.

adriancole:
Game of Thrones has an http client? awesome..

juhovh:
@jcchavezs I have removed all changes to the yarn.lock file as requested. I must say running yarn install will re-add them there though, but they are not necessary for this to work.

@adriancole Well, I guess the winter is coming, and they need all the publicity they can get for the final season. 😜

Issue #348, opened 2019-03-26
Add axios instrumentation


adriancole:
There are a number of broken axios instrumentation out there. Maybe it is time to have a first party one that has all the tests running etc? I'm not sure on this, so deferring to other maintainers to make the call.

@thegryfon has been working on fixups to https://github.com/geekeren/zipkin-js-instrumentation-axios 

@geekeren would you be interested in contributing yours?

geekeren:
@adriancole Yes, I am glad to contribute my code, what can I do?

adriancole:
@geekeren great! I think best to make a directory like we have here in a pull request. make sure you carry over your good README content and double-check the integration tests are latest.

https://github.com/openzipkin/zipkin-js/tree/master/packages

thegryfon:
Hi @adriancole & @geekeren , so some discussion topics that I would like to propose regarding this enhancement :
1. target environment for running this instrumentation (nodejs apps,browser apps)
1. the ideal way to leverage tracer.scoped() statement in order to contain trace id context (either put it explicitly in the documentation how to, or transparently in interceptors) and the pros / cons
1. adjust the unit test, so that for each unit test sequence a new axios instance will be created, otherwise when each request creates new id,  the assertions might get mixed up with previous unit test blocks and become failed. The reason for this the : wrapAxios is called repeatedly on the same axios instance, so that actually axios interceptor has multiple instances. see for example (https://github.com/geekeren/zipkin-js-instrumentation-axios/blob/45ec6525cb98968566801a27e7fc52a33a024f51/test/integrationTest.js#L50) and ( https://github.com/geekeren/zipkin-js-instrumentation-axios/blob/45ec6525cb98968566801a27e7fc52a33a024f51/test/integrationTest.js#L91 ). Because there are multiple instance of interceptors, instrumentation.recordRequest will be called twice, and as you know each time recordRequest is called, it will create a different id. Current integration test will not fail because there are no scoped request in the test, hence traceId will always be the same. 
[...]

PR #356, opened 2019-04-24
#348 pass traceId from request to response

geekeren:
- pass traceId from request to response by config, traceId will change after record request
- avoid dirty `axios`, create an axios instance before using interceptors

### DIFF
--- a/packages/zipkin-instrumentation-axios/src/index.js
+++ b/packages/zipkin-instrumentation-axios/src/index.js
@@ -3,7 +3,20 @@ import {Instrumentation} from 'zipkin';
 export const wrapAxios = (axios, options = {}) => {
   const {tracer} = options;
   const instrumentation = new Instrumentation.HttpClient(options);
-  const zipkinRecordError = error => {
+  const zipkinRecordRequest = config => tracer.scoped(() => {
+    const newConfig = instrumentation.recordRequest(
+        config,
+        config.url,
+        config.method
+      );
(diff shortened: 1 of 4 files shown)


adriancole:
relevant discussion https://github.com/openzipkin/brave/issues/892

jcchavezs:
@geekeren thanks for this! Would you add also a test case for covering this change?

geekeren:
@jcchavezs  Done

jcchavezs:
Ping @thegryfon. I haven't seen this feature (of passing the traceId in the response) anywhere else, not sure, I would like to have some input around this from @adriancole

packages/zipkin-instrumentation-axios/src/index.js
```diff
@@ -1,28 +1,36 @@
...
+    const newConfig = instrumentation.recordRequest(
+        config,
+        config.url,
+        config.method
+      );
+    newConfig.traceId = tracer.id;
+    return newConfig;
+  });
+  const zipkinRecordResponse = res => tracer.scoped(() => {
+    instrumentation.recordResponse(res.config.traceId, res.status);
```
geekeren: @jcchavezs If here I use `tracer.id` instead of `res.config.traceId`, the tests will fail. I don't know why and I think I don't understand zipkin-js well.

adriancole:
merging for the sake of progress. thanks!

PR #357, opened 2019-04-25
Add Zipkin instumentation for kafka messages through KafkaJS

high-stakes:


jcchavezs:
Thanks a lot @highstakes we will review this carefully. Ping @ghermeto @jeqo

tor. 25. apr. 2019, 19:58 skrev highstakes <notifications@github.com>:

jcchavezs:
Some insights on the questions:


> how to do testing in this repo? since Kafka headers are required we need kafka version 0.11+ (not supported by the kafka-please node dev dependency). Should I test against a live kafka instance or mocks? In case of a real instance, is using docker compose okay?
I run some tests against kafka in https://github.com/openzipkin/zipkin-php-example/tree/kafka_transport (see https://github.com/openzipkin/zipkin-php-example/blob/kafka_transport/.circleci/config.yml and https://github.com/openzipkin/zipkin-php-example/blob/kafka_transport/docker-compose.yml), the average build did not take more than a minute (using docker-compose). i believe it is worth to try and if it has a huge impact we could think in something else. Something to consider here is that if we include a real kafka in tests, we should make sure test are not flickering (i.e. make sure tests runs after kafka is available). @jeqo is a great kafka expert if you have questions.

> error handling is supported but if the user doesn't handle it, retrial will keep happening forever and append further error spans to the trace. My preferred solution is to allow only sending one error and stop sending further errors per for that topic until it recovers. The only problem is that I did not find support for throwing away a partial span, is there any way to do this with zipkinjs? (I know the java implementation supports it)

If the span is finished then it goes to reporting. As for retries, better to get @jeqo or @adriancole for some input.

high-stakes:
Thanks @jcchavezs , I will get started with that. About the span reporting, I meant the problem is when I start a span for example when message consumption starts and there is an error I would need to drop the existing span

jcchavezs:
@high-stakes right. Right, so I faced a similar thing with zipkin-go where there is not way to abandom an span. What we do there is basically I create the span after I know all the information on before hand: https://github.com/jcchavezs/zipkin-instrumentation-sql/blob/master/driver.go#L128 however I could not find a way in zipkin-js to pass the start time into the span so I just created an issue for that cc @ghermeto

high-stakes:
@jcchavezs that workaround would prevent users to add extra tags to the active span, no? I think this feature could be supported in the core zipkin module to avoid workarounds (similar to brave with finishedspanhandler), what do you think? 
Other than that, I have added finalized the instrumentation (no automatic error handling for now) and added tests, please let me know what you think.

jcchavezs:
@high-stakes totally. I think to abandom a span should be a thing. Could you please fill in an issue for that? In the meanwhile I will merge this PR.

PR #363, opened 2019-05-27
Updating Timestamp in Console Recorder

gaonkar-adi:
Adding Timestamp to the Console Recorder Logger to record time across multiple requests.


jcchavezs:
@gaonkar-adi thanks for coming up with this. It looks great to me but let's keep it for 24 hours so someone else could raise concerns.

gaonkar-adi:
@jcchavezs Thanks for the feedback. Sure, I can take it up if there are any concerns.

PR #364, opened 2019-06-11
Reflect supportsJoin agument parameter in types

SerCeMan:
As far as I can see from [index.js](https://github.com/openzipkin/zipkin-js/blob/00604b73d938b25b36a754765cdcca435def1b07/packages/zipkin/src/tracer/index.js#L31), `supportsJoin` is a valid argument, so this PR reflects this in the typed arguments. 

Thank you for the awesome library!


jcchavezs:
Hi @SerCeMan thanks for this!

SerCeMan:
Wow, that was a swift turnaround, thanks!

PR #367, opened 2019-06-26
Refactors TraceId type against known problems; fixes shared flag

adriancole:
Before, the shared flag wasn't implemented correctly, which could lead
to data problems. Moreover, `TraceId.parentId` returned a heuristic
of the same as `TraceId.spanId` when absent: this causes data problems
and is an known antipattern. Finally, 'flags', an implementation detail
of the old B3 header was exposed instead of 'debug'

This fixes all of these problems. Another commit will make the change
compatible with existing implementations via deprecations:

* `TraceId.parentId` -> `TraceId.parentSpanId` with option result
* `TraceId.flags` -> `TraceId.isDebug()` -- already present
* Bad `TraceId` ctor parameter which allowed `traceId` to be absent: conditional detection, favoring plain str
ing.

### DIFF
--- a/packages/zipkin-encoder-thrift/test/index.test.js
+++ b/packages/zipkin-encoder-thrift/test/index.test.js
@@ -81,7 +81,7 @@ describe('Thrift v1 Formatting', () => {
     const span = new Span(new TraceId({
       traceId: '000000000000162e',
       spanId: '000000000000162e',
-      debug: 1
+      debug: true
     }));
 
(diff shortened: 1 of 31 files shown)


adriancole:
ok ready to go

adriancole:
after this is merged, I'll do more cleanups, notably change integration tests to assert in v2 format as the current annotation based assertions obscure impact

PR #369, opened 2019-06-27
Refactors Axios unit tests

adriancole:
Step 1: make the request headers test do what it says it does

### DIFF
--- a/packages/zipkin-instrumentation-axiosjs/test/integrationTest.js
+++ b/packages/zipkin-instrumentation-axiosjs/test/integrationTest.js
@@ -14,6 +14,7 @@ describe('axios instrumentation - integration test', () => {
   let apiPort;
   let record;
   let tracer;
+
   before((done) => {
     mockServer().then(server => {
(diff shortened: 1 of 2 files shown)


packages/zipkin-instrumentation-axiosjs/test/integrationTest.js
```diff
@@ -22,53 +23,41 @@ describe('axios instrumentation - integration test', () => {
...
   });
+
+  const getClient = () => wrapAxios(axios, {tracer, serviceName, remoteServiceName});
+
   it('should add headers to requests', done => {
     tracer.scoped(() => {
       const zipkinAxiosClient = getClient();
       const urlPath = '/weather/wuhan';
       const url = `http://${apiHost}:${apiPort}${urlPath}?index=10&count=300`;
-      zipkinAxiosClient
```
adriancole: @openzipkin/zipkin-js-champions in I think every integration test, we copy/pasted the same code which didn't actually test that headers were added!

PR #377, opened 2019-06-28
fix: checks if there is a start timestamp to properly detect timeout.

jcchavezs:
If a span has no startTimestamp timeout checks should not apply.

Problem with spans with no timestamp is that the UI can't render it.

Ping @adriancole

### DIFF
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -96,7 +96,7 @@ class BatchRecorder {
 
     // TODO(adriancole) refactor so this responsibility isn't in writeSpan
     if (span === undefined) {
-      // Span not found.  Could have been expired.
+      // Span not found. Could have been expired.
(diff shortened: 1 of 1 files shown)


adriancole:
needs a test

adriancole:
also we do want to flush data regardless of whether there was a timestamp as this could be a bug in instrumentation or late data. lack of timestamp isn't a good enough signal to do anything except possibly not calculate against it. Inside the BatchRecorder, there's separate bookkeeping on the PartialSpan which is always assigned a timestamp... from this diff I can't tell which type of span was being referred to. Basically I think this change should be reverted or backfilled with a test rather than add guards which may be distracting

```
class PartialSpan {
  /**
   * @constructor
   * @param {TraceId} traceId
   */
  constructor(traceId) {
    this.traceId = traceId;
    this.startTimestamp = now();

```

PR #378, opened 2019-06-28
Backfills integrated tests of BatchRecorder

adriancole:
There were some red herrings about an example failing. This carves out
a unit test for people to fill for suspected problems.

### DIFF
--- a/packages/zipkin/test/batch-recorder.integrationTest.js
+++ b/packages/zipkin/test/batch-recorder.integrationTest.js
@@ -37,6 +37,7 @@ describe('Batch Recorder - integration test', () => {
       kind: 'CLIENT',
     });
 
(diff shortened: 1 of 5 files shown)


packages/zipkin/test/batch-recorder.integrationTest.js
```diff
@@ -0,0 +1,86 @@
...
+
+    expectSpan(popSpan(), {
+      kind: 'PRODUCER',
+    });
+
+    expect(recorder.partialSpans).to.be.empty; // eslint-disable-line no-unused-expressions
+  });
+
+  it('should start and finish a consumer span', () => {
+    recorder.record(new Record({traceId, annotation: new Annotation.ConsumerStart()}));
```
adriancole: one thing I've struggled to assert is that recorder.partialSpans should have contents at this point. I can't get a chai assertion to show it.

PR #383, opened 2019-06-29
Fixes various timestamp problems, notably when a span is flushed

adriancole:
Before, we conflated timing around data timeouts with span lifecycle
(eg start/finish). This caused some confusion in the code, and a lot of
redundant clock calls. This refactors the code to only use timestamps
from the events, and only the authoritative ones for purposes of start
and finish.

Moreover, this attempts to compensate for data lost when a flush happens
between start and finish by backfilling a "finish" annotation in such
case. With better code structure in finish callbacks, we should even be
able to eliminate this (by redundantly replaying the span's start event
on an async finish callback). For now, it will correct the problems of
flush better than most libraries do, because the UI will interpret any
annotation to backfill a duration.

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -112,6 +112,11 @@ class Tracer {
     return childId;
   }
 
+  // this allows you to avoid use of implicit trace ID and defer implicit timestamp derivation
+  _explicitRecord(traceId, annotation, timestamp = now(this._startTimestamp, this._startTick)) {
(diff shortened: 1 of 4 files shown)


packages/zipkin/src/tracer/index.js
```diff
@@ -196,16 +196,10 @@ class Tracer {
     return this._localEndpoint;
   }
 
-  recordAnnotation(annotation, timestamp = now(this._startTimestamp, this._startTick)) {
```
adriancole: before, even when we didn't sample, we incurred a timestamp cost. now, it is deferred

Issue #384, opened 2019-06-30
Finish kafka-js and knock out kinks


adriancole:
very cool that @high-stakes got kafka started. Upon inspection I noticed that we aren't modeling this consistently with the brave instrumentation. Notably, there are some batch concerns and also spans are re-using the producer id instead of forking a child. The latter is a problem as multiple consumers can consume the same message which will screw up data if they report into the same ID.

I'll work on fixing this up as a general part of preliminary work needed before we expose a v2 api. cc @ImFlog @jeqo

adriancole:
related: pinged slack to see if kafkajs is known to work in browser env or not https://kafkajs.slack.com/archives/CF6RFPF6K/p1561867307001500

bbarmin:
Thanks @adriancole , great changes. The open questions are definitely the batch handling and the error handling. When production or consumption fails KafkaJS currently will keep retrying with exponential fallback unless there is logic added to prevent this. I am wondering if this should be prevented (In order to not keep adding spans to the trace forever). I was thinking of adding the 'finished span handler' functionality to zipkinjs as a potential fix to limit reporting on a topic basis until recovery from error and only report the first x error spans. #359

adriancole:
kafkajs won't work on browser. Another TODO here is to refactor the tests as currently they take over half the time of running all tests including typescript. (>30s out of 55s for the entire project)

PR #385, opened 2019-06-30
Refactors kafkajs tests so that instrumentation errors are easier to see

adriancole:
This refactors the tests for kafkajs to use v2 format and fixes some lower hanging problems

* docker containers didn't cleanup which made debugging mysterious
* setup was being used same way as client-server, which is incorrect
  * remote service should be the broker, not the guessed name of a server
  * producer and consumer should never share span ID (join) especially because this library does retries for failed message processors!
* incorrectly read Buffer as a string which broke propagation
* incorrectly set remote service name where service name goes
* mislabled span names differently than the calling functions

Notable concerns left out of scope are:
* consumer span is being labeled on a processor function, which is a modeling glitch. Consumer should represent the time spent getting the message from kafka, which is quite short.
* docker setup is ok now, but still could be better possibly if managed by mocha somehow instead of shell script

See #384


adriancole:
this still needs work to do as there's a bunch of timeout problems in both kafka via docker, and also the sequencing of promises in general. I suspect we should be looking at testcontainers-node.. cc @bsideup in case you know an existing setup

packages/zipkin-instrumentation-kafkajs/src/kafka-recorder.js
```diff
@@ -12,10 +16,10 @@ const recordConsumeStart = (tracer, {topic, partition, message}) => {
 
     // TODO: this should definitely note join. It should make a child
     id = tracer.join(new TraceId({
-      traceId,
-      parentId: fromNullable(parentId),
-      spanId,
-      sampled: fromNullable(sampled),
+      traceId: bufferToAscii(traceId),
```
adriancole: we might consider moving this logic into traceId, if it is common for headers to be of type Buffer, though I think Buffer might be node.js only and so far this is the only user of it.

adriancole:
ps the timeouts are not so much a problem of tests not running (they do) it is more about the obfuscation of issues. It isn't as high priority to manage test differently, just it is hard to maintain especially the test that looks for errors to occur.

packages/zipkin-instrumentation-kafkajs/package.json
```diff
@@ -5,18 +5,19 @@
   "main": "lib/zipkin-instrumentation-kafkajs.js",
   "scripts": {
     "build": "babel src -d lib",
-    "pretest": "docker-compose up -d; timeout 60 bash -c 'until echo > /dev/tcp/localhost/9092; do sleep 1; done'",
-    "test": "mocha --require ../../test/helper.js",
+    "pretest": "(docker kill kafkajs-test; docker rm kafkajs-test) 2>&-; docker run -d -p 9092:9092 --name kafkajs-test --hostname localhost openzipkin/zipkin-kafka && timeout 60 bash -c 'until echo > /dev/tcp/localhost/9092; do sleep 1; done'",
```
adriancole: in local dev, tests would fail because the pre-existing docker containers weren't deleted. the result would be strange as it would appear parsing was busted because the same headers are always read... this was due to replaying whatever the first test message was. now this is fixed, but really we have to switch this to something else!

bsideup:
@adriancole one can run the tests with GraalVM and enjoy the Java version of Testcontainers :D
https://medium.com/graalvm/using-testcontainers-from-a-node-js-application-3aa2273bf3bb

adriancole:
OK I rewrote the description. I'm done with this work because there are a lot more modules to refactor and the most important things here are addressed now.

packages/zipkin-instrumentation-kafkajs/README.md
```diff
@@ -24,8 +24,8 @@ const kafka = instrumentKafkaJs(new Kafka({
   clientId: 'my-app',
   brokers: ['localhost:9092']
 }), {
-    tracer, //Your zipkin tracer instance
-    remoteServiceName : 'node-backend' //When using producer, the remote endpoint
+    tracer, // Your zipkin tracer instance
+    remoteServiceName : 'kaka' // This should be the symbolic name of the broker, not a consumer.
```
jcchavezs: is `kaka` correct?

bsideup: given that "kaka" in Russian means "poop", that's a very rhetorical question, although I'm sure Adrian likes Kafka and that's just a typo :D

adriancole: LOL actually I misspelled.. I meant caca or more specifically 💩

PR #389, opened 2019-07-02
Stops encouraging serviceName overrides

adriancole:
We continue to get support problems due to people overriding service
names incorrectly. There's rarely a valid reason to override what's in
the tracer. This stops us from doing it in tests and README files. We
should strongly consider removing the ability to override the
serviceName in v2 such as we do in other libraries, for the reason of
needless support burden.


adriancole:
Example recent problem while mentioned about finagle, it is the same situation here. https://groups.google.com/forum/#!topic/zipkin-user/EgTN97wO4Sk

In the future, we can add a feature to override service names consistently, if someone is sharding a process into endpoint-scoped services (using local root id to ensure the data change is scoped and propagated downwards). For now, it is best really to stop tempting people to change this by showing examples!

PR #392, opened 2019-07-05
Removes lodash to avoid distracting bot pull requests

adriancole:
We have almost no usage of lodash, and the usage we have isn't security
critical. This removes the lodash dependency to avoid having to look up
what unrelated problems exist when there's a security notice.

See #390
Fixes #391

### DIFF
--- a/packages/zipkin-instrumentation-request-promise/src/request.js
+++ b/packages/zipkin-instrumentation-request-promise/src/request.js
@@ -1,5 +1,4 @@
 import request from 'request-promise';
-import _ from 'lodash';
 import {Instrumentation} from 'zipkin';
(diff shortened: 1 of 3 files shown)


packages/zipkin-instrumentation-request-promise/src/request.js
```diff
@@ -92,13 +91,13 @@ const Request = class {
      * breaks zipkin scope
      */
     let cb = callback;
-    if (_.isFunction(callback)) {
+    if (typeof callback === 'function') {
```
adriancole: all of our other code uses this form. If we need a utility more robust, I think we should vendor the 4 lines

PR #393, opened 2019-07-06
Migrates fetch to v2 style assertions and ensures they run in a browser

adriancole:
This migrates the same tests used in Axios to fetch, notably deferring
loading of node-fetch in case we are running in a browser.


adriancole:
note I'm aware this is largely copy/paste from axios tests. I plan to refactor a base test fixture once all of the clients have tests backfilled. Guessing what things will look like ahead of time would instead cause drift I don't feel like dealing with right now.

PR #394, opened 2019-07-06
Fixes bug where Tracer.id sentinel value was ignored

adriancole:
The initial design of zipkin-js included a sentinel value on `Tracer.id`
that allows you to avoid needing to compare against null. The trouble is
that the sentinel was never compared against, and would result in root
spans appearing to be children of the sentinel. This is a bug because it
can lead to very large pseudo-traces. I'm sure the intent was to check
the sentinel before using it.

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -41,11 +41,13 @@ declare namespace zipkin {
       log?: Console
     });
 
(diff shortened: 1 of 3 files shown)


adriancole:
I found this while looking at the got instrumentation tests.. basically root got client requests were showing up as children of the fake ID.

PR #399, opened 2019-07-07
Fixes double-recording bug in superagent and rewrites tests to v2

adriancole:
We were calling finish twice on 4xx and 5xx and didn't notice.


packages/zipkin-instrumentation-superagent/src/superagentPlugin.js
```diff
@@ -19,26 +19,38 @@ const {Instrumentation} = require('zipkin');
...
-    const recordResponse = (res) => {
+    // capture the trace ID from recording the start of the request
+    let traceId;
+    const {headers} = tracer.scoped(() => {
+      const injected = instrumentation.recordRequest(req, url, method);
+      traceId = tracer.id;
+      return injected;
+    });
+
+    // We can't use end() as the caller might be using it.
```
adriancole: we should probably do another cleanup pass later, if we can determine how to get the response and error callbacks at the same time without using end()

PR #406, opened 2019-07-08
refactors connect integration tests

adriancole:


adriancole:
Dependent on #407 as there was an eslint glitch that went away in v5

adriancole:
finally :P

PR #410, opened 2019-07-09
Converts express proxy tests to v2 format

adriancole:


jcchavezs:
Not sure about what "better error message on 500" means. I think the status
code on error tags are quite standard a among http instrumentations and I
would keep it as it is. Apologies if that is not what you meant in the TODO.

tir. 9. jul. 2019, 16:06 skrev Adrian Cole <notifications@github.com>:

adriancole:
@jcchavezs the error tag is intended to be a real error. duplicating the status code is a fallback. In a lot of cases, there's a better message available and we should certainly use it.

adriancole:
the main problem is how the middleware is written now, it doesn't include an error callback. An example of a better error tag is the reason why the 500 happened, ex a downstream proxy failure. 500 doesn't tell you much eventhough it is important to keep the tag 500 to know that was the status returned.

adriancole:
ex in redis we have better

```javascript
  if (error) tracer.recordBinary('error', error.message || String(error));
```

This allows us to see the actual error in the middleware, such as
```
  error: 'ERR wrong number of arguments for \'get\' command'
```

Make sense?

adriancole:
to finish this statement basically most http middleware are presumably accidentally not passing error through, which means application exception in middleware doesn't take like it is written to.

```javascript
  recordResponse(id, statusCode, error) {
    this.tracer.setId(id);
    this.tracer.recordBinary('http.status_code', statusCode.toString());
    if (error) {
      this.tracer.recordBinary('error', error.toString());
    } else if (statusCode < 200 || statusCode > 399) {
      this.tracer.recordBinary('error', statusCode.toString());
    }
    this.tracer.recordAnnotation(new Annotation.ServerSend());
}
```

https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/instrumentation/httpServer.js#L84-L94

jcchavezs:
Right so I think I misunderstood the TODO. I agree when there is a better error we can provide then it is good to use that instead of status code. What I thought you were proposing was to translate the status code into something `human readable` like I reverted in https://github.com/jcchavezs/zipkin-instrumentation-symfony/pull/32/files#diff-2f9ab9d511547bb0d187a3aafe9c03c5L126. Basically if the actual library provides something meaningful I agree we should use it, my concern was more like converting `500` into `Internal Server Error` which I think is not your intention here.

adriancole:
yeah I understand your concern and agree with it. definitely not aiming for
that. thanks for keeping watch!

PR #412, opened 2019-07-09
Fixes hapi trace scoping and ports tests to v2

adriancole:
Before, the current trace ID was not available during the route handler method. This would cause outgoing requests to be a part of the wrong trace.

### DIFF
--- a/packages/zipkin-instrumentation-hapi/test/integrationTest.js
+++ b/packages/zipkin-instrumentation-hapi/test/integrationTest.js
@@ -13,6 +13,7 @@ describe('hapi instrumentation - integration test', () => {
 
   let spans;
   let tracer;
+  let sentinelTraceId;
 
   beforeEach(() => {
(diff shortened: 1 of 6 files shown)


packages/zipkin-instrumentation-hapi/test/integrationTest.js
```diff
@@ -1,185 +1,195 @@
...
+
+  function successSpan(path, city) { // eslint-disable-line no-unused-vars
+    return ({
+      name: 'get',
+      kind: 'SERVER',
+      localEndpoint: {serviceName, ipv4},
+      tags: {
+        'http.path': path,
+        'http.status_code': '200',
+        // city TODO: scoping for the handler function
```
adriancole: ex test found that it is very likely downstream requests will break. if you cannot see the trace id to add a tag, it is likely you also cannot make a child span

PR #416, opened 2019-07-19
feat(#381): adds zipkin-js.flush annotation for when a span is report…

jcchavezs:
This PR makes it possible to add a `zipkin-js.flush` annotation for a span when it is reported due to timeout (i.e. the span has been notified but the finish method was not called).

Closes #381 

Ping @adriancole

PR #417, opened 2019-07-20
feat(#362): adds flush method to http reporter.

### LINKED ISSUE #362: Add flush to HttpLogger
I want to use zipkin-js to trace requests handled by an AWS lambda function, written in NodeJS but the lambda seems to be dropping the request context prior to the async POST call to my Zipkin collector.

Currently, I'm using the ExplicitContext attribute from zipkin-js and the lambda function dies before the reporter can send the span.

I suspect the AWS Lambda drops the request context immediately after receiving the response, and since the reporter sends the POST request to the collector on a 1 sec cadence, the reporter is unable to send the span data.

A possible resolution may be to make the reporter synchronous (i.e: block the thread until the span is reported, and only then finish the request).

And another possible resolution may be to add a flush option, which would ideally asynchronously close the POST http request to the Zipkin collector, prior to finishing the request and the lambda shutting down.

Any advice is appreciated 👍

jcchavezs:
This PR adds a `flush` method for http reporter.

Closes #362 

Ping @CSP197 @adriancole

### DIFF
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -8,12 +8,6 @@ const {Span, Endpoint} = require('./model');
 const defaultTimeout = 60 * 1000000;
 
 /**
- * default batch interval 1 second (in miliseconds)
- * @type {number}
- */
-const defaultBatchInterval = 1000;
-
-/**
  * defaultTags property name
  * @type {symbol}
  */
@@ -69,18 +63,12 @@ class BatchRecorder {
    * @param {Object} options
    * @property {Logger} logger logs the data to zipkin server
    * @property {number} timeout timeout for span in microseconds
-   * @property {number} batchInterval interval for reporting in miliseconds.
-   * The value -1 means that there won't be batch reporting on interval basis.
    */
-  constructor({logger, timeout = defaultTimeout, batchInterval = defaultBatchInterval}) {
+  constructor({logger, timeout = defaultTimeout}) {
     this.logger = logger;
     this.timeout = timeout;
     this.partialSpans = new Map();
     this[defaultTagsSymbol] = {};
-
-    if (batchInterval === -1) {
-      return;
-    }
 
     // read through the partials spans regularly
     // and collect any timed-out ones
@@ -94,7 +82,7 @@ class BatchRecorder {
           this._writeSpan(id, span);
         }
       });
-    }, batchInterval);
+    }, 1000);
     if (timer.unref) { // unref might not be available in browsers
       timer.unref(); // Allows Node to terminate instead of blocking on timer
     }
@@ -147,6 +135,9 @@ class BatchRecorder {
     }
   }
 
+  /**
+   * Calling this will flush any pending spans to the transport.
+   */
   flush() {
     this.partialSpans.forEach((span, id) => {
       this._writeSpan(id, span);
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -8,6 +8,12 @@ const {Span, Endpoint} = require('./model');
 const defaultTimeout = 60 * 1000000;
 
 /**
+ * default batch interval 1 second (in miliseconds)
+ * @type {number}
+ */
+const defaultBatchInterval = 1000;
+
+/**
  * defaultTags property name
  * @type {symbol}
  */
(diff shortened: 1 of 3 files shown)


packages/zipkin/src/batch-recorder.js
```diff
@@ -135,6 +147,15 @@ class BatchRecorder {
     }
   }
 
+  flush() {
+    this.partialSpans.forEach((span, id) => {
+      if (span.delegate.duration === undefined) {
```
jcchavezs: if the span is finished, we don't want to add akward annotations outside the span lifespan. Do you agree?

adriancole:
actually I am not sure the logic here is timeout or interval in nature. my recommendation is to remove extraction of the time parameter and just expose flush. that way we dont have to setup a new parameter just to remove it later as we need to redo the batching part anyway.

jcchavezs:
Makes sense both comments. It actually makes things easier and actually the
go reporter works the same way as you describe

søn. 21. jul. 2019, 06:41 skrev Adrian Cole <notifications@github.com>:

jcchavezs:
PTAL

jcchavezs:
PTAL

søn. 21. jul. 2019, 11:32 skrev José Carlos Chávez <jcchavezs@gmail.com>:

packages/zipkin/test/batch-recorder.test.js
```diff
@@ -223,6 +223,39 @@ describe('Batch Recorder', () => {
...
+      trace.recordAnnotation(new Annotation.ServerRecv());
+    });
+
+    clock.tick(100); // 1000 is de batching interval
+
+    expect(logSpan.calledOnce).to.equal(false);
+
+    recorder.flush();
+
+    expect(logSpan.calledOnce).to.equal(true);
```
adriancole: maybe you can rebase and redo this test. A similar old one was lying.

packages/zipkin/src/batch-recorder.js
```diff
@@ -135,6 +135,15 @@ class BatchRecorder {
     }
   }
 
+  /**
+   * Calling this will flush any pending spans to the transport.
```
adriancole:
```suggestion
   * Calling this will flush any pending spans to the transport.
   *
   * Note: the transport itself may be batching, in such case you may need to flush that also.
```

packages/zipkin/test/batch-recorder.test.js
```diff
@@ -302,8 +302,35 @@ describe('Batch Recorder - integration test', () => {
...
+          spans.push(JSON.parse(JSON_V2.encode(span)));
+        }
+      }
+    });
+
+    recorder.record(record(rootId, now(), new Annotation.ServerRecv()));
 
     expect(spans).to.be.empty; // eslint-disable-line no-unused-expressions
+
+    clock.tick(100); // 1000 is de batching interval
```
adriancole: I would personally remove the lolex stuff as it is irrelevant. For example, we have other tests in the same file which show that records are not flushed until complete. The only responsibility this one should have is that when you hit flush the data is expected. The clock ticking stuff distracts from an otherwise simple goal IOTW. For this test you can even use fake timestamps like other things here

packages/zipkin/test/batch-recorder.test.js
```diff
@@ -302,8 +302,35 @@ describe('Batch Recorder - integration test', () => {
...
+
+    expect(spans).to.be.empty; // eslint-disable-line no-unused-expressions
+
+    recorder.flush();
+
+    expect(popSpan()).to.deep.equal({
+      traceId: rootId.traceId,
+      id: rootId.spanId,
+      kind: 'SERVER'
```
adriancole: surprised the timestamp didn't make it!

packages/zipkin/test/batch-recorder.test.js
jcchavezs: Right. I think for me it was better to make it clear that the flush occurs when it was not yet the time for a scheduled flush.

packages/zipkin/test/batch-recorder.test.js
jcchavezs: I guess it was because of `lolex`. I think this is ready to be merged.

adriancole:
win!

adriancole:
@jeffthompson1971 what you are describing is a really bad bug it seems. I've tried to help you on gitter before and maybe notifications aren't working. there's a human communication problem by which I'm not able to get a suggestion to you to try connect instead (ideally avoiding the bug) https://gitter.im/openzipkin/zipkin?at=5d93e7a9fb131014e721b07f

At any rate, you can mask the bug by setting the batch recorder timeout to a very high number, which essentially makes it impossible to ever become timed out.

adriancole:
note that if you do set this to a really high number there's a chance, depending on the nature of the bug, that it could leak memory. I really implore trying some suggestions to make the bug not happen

jcchavezs:
Hi @jeffthompson1971,

What you are describing is really bad but I am not sure this is a problem caused for this PR. This PR includes a method for flush but that method is being called by the user, not by any zipkin instrumentation. The piece of code that *adds the `zipkin-js.flush` annotations* is https://github.com/openzipkin/zipkin-js/pull/416 and that is adding the annotation. What I think you are experiencing is that the middleware is holding the span and not finishing. Could you please provide your setup over gitter?

jcchavezs:
Hi @jeffthompson1971 we just released https://github.com/openzipkin/zipkin-js/releases/tag/v0.19.1. This should fix the problem you have in express as hooks usage have been fixed, would you mind trying it?

jeffthompson1971:
sorry I missed this guys.  I will try this today!

jeffthompson1971:
Actually I already tried it I had forgotten. Still see the issue.  And I was also given the recommendation to move to zipkin-instrumentation-connect and I did that too.. still i get these zipkin.flush spans.  wish those would just go away until i figure out why this is happening.  Just not easy to debug this it seems.

PR #418, opened 2019-07-22
ports core library test assertions to v2 and fixes a couple bugs

adriancole:
took all my battery!


packages/zipkin/src/batch-recorder.js
```diff
@@ -154,7 +162,7 @@ class BatchRecorder {
           break;
         case 'ServerRecv':
           span.delegate.setShared(id.isShared());
-          span.delegate.setKind('CLIENT');
+          span.delegate.setKind('SERVER');
```
adriancole: oops!

PR #419, opened 2019-07-23
Uses mocha shared behaviors to setup client integration test middleware

adriancole:
This starts to pare down test duplication by extracting out a facet that
manages the start and stop of middleware used in client integration
tests.

Installation should be like this
```javascript
const {setupTestServer} = require('../../../test/testFixture');

const server = setupTestServer();
```

Later, you can get a url for clients to use via `server.url(path)`

See https://github.com/mochajs/mocha/wiki/Shared-Behaviours


jcchavezs:
+1

adriancole:
polished for merge

PR #420, opened 2019-07-24
Extracts a tracer fixture to remove copy/paste and fixes scoping problems in instrumentation

adriancole:
This ports all the instrumentation packages. There is still some porting opportunities in the reporters and the base module itself. However, this is a good place to stop for a bit.

### DIFF
--- a/packages/zipkin-instrumentation-connect/test/integrationTest.js
+++ b/packages/zipkin-instrumentation-connect/test/integrationTest.js
@@ -1,43 +1,28 @@
 const {expect} = require('chai');
-const {ExplicitContext, InetAddress, Tracer} = require('zipkin');
-
+const {InetAddress} = require('zipkin');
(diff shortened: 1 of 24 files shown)


adriancole:
will try to finish this today

packages/zipkin-instrumentation-connect/test/integrationTest.js
```diff
@@ -67,26 +52,32 @@ describe('connect instrumentation - integration test', () => {
   }
 
   describe('restify middleware', () => {
+    // Restify uses async hooks. Until there is a CLS hooked implementation here, we need to be
```
adriancole: this was was an oldie but goodie heh. once things are stable we really need that cls hooked thing merged and released!

adriancole:
OK I am done with this PR

PR #423, opened 2019-07-24
Loosens type checking during join as it can mismatch due to transpilation

### LINKED ISSUE #422: Error: Must be valid TraceId instance
After upgrading to version 0.18.5 of zipkin-instrumentation-restify and zipkin, I am getting the following error when the request to my server comes in with a traceId: "Error: Must be valid TraceId instance".

I've narrowed the problem down to the following snippet of code in [zipkin/tracer/index.js](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/tracer/index.js#L171) inside the join function

```
if (!(traceId instanceof TraceId)) {
    throw new Error('Must be valid TraceId instance');
}
```

This if is always resulting to false, after inspecting the traceid passed to this function, I found the following.

traceId parameter:
```
TraceId {
  _traceId: 'b7daf2ae1b9e2555',
  _parentId:
   { type: [Getter],
     present: [Getter],
     map: [Function: map],
     ifPresent: [Function: ifPresent],
     flatMap: [Function: flatMap],
     getOrElse: [Function: getOrElse],
     equals: [Function: equals],
     toString: [Function: toString] },
  _spanId: 'b7daf2ae1b9e2555',
  _sampled:
   { type: [Getter],
     present: [Getter],
     map: [Function: map],
     ifPresent: [Function: ifPresent],
     flatMap: [Function: flatMap],
     getOrElse: [Function: getOrElse],
     equals: [Function: equals],
     toString: [Function: toString] },
  _debug: false,
  _shared: false }
```

`traceId instanceof TraceId_1`: false
`traceId.constructor.name`: TraceId
`TraceId_1.constructor.name`: Function
`TraceId.constructor.name`: Function

Please note: The usage of TraceId_1 is because that is the output produced after transpilation, and hence is what is being checked in the code when it is actually running.

```
if (!(traceId instanceof TraceId_1)) {
    throw new Error('Must be valid TraceId instance');
}
```

This issue only happens if the request is made with the traceid and spanid headers populated.

adriancole:
Fixes #422

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -168,9 +168,16 @@ class Tracer {
   }
 
   join(traceId) {
-    if (!(traceId instanceof TraceId)) {
+    if (isUndefinedOrNull(traceId)) {
+      throw new Error('traceId is a required arg');
+    }
+
+    // duck type check until we sort out a better way. We don't want to break
+    // transpiled usage ex. `traceId instanceof TraceId_1: false` See #422
+    if (isUndefinedOrNull(traceId._spanId)) {
       throw new Error('Must be valid TraceId instance');
     }
+
     if (!this.supportsJoin) {
       return this.createChildId(traceId);
     }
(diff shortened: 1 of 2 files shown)


adriancole:
will take any feedback post merge. need to get folks running again!

PR #425, opened 2019-07-25
Plugs context leak when something throws inside a scoped block

adriancole:


adriancole:
This only impacted ExplicitContext

PR #427, opened 2019-07-26
Creates common fixture for http client test

adriancole:
By using a common http client test fixture, we can remove needless difference between instrumentation tests. This makes it easy to spot and correct missing tests, and also easier to identify special behaviors.


adriancole:
fixed several CujoJS bugs despite it being an abandoned library..

jcchavezs:
Wow, this is such a great simplification.

fre. 26. jul. 2019, 09:36 skrev Adrian Cole <notifications@github.com>:

adriancole:
finally.. loc count going down

PR #428, opened 2019-07-26
Update TypeScript type definitions for Tracer class

Igor-lkm:


packages/zipkin/index.d.ts
```diff
@@ -49,7 +49,7 @@ declare namespace zipkin {
     local<V>(name: string, callback: () => V): V;
     createRootId(isSampled?: option.IOption<boolean>, isDebug?: boolean): TraceId;
     /** Creates a child of the current trace ID or a new root span. */
-    createChildId(): TraceId;
+    createChildId(parentId: TraceId): TraceId;
```
adriancole: is there a typescript note that indicates the parameter is optional?

ps probably good to also add a test that uses this. thanks for the help

PR #429, opened 2019-07-27
Reduces unnecessary use of beforeEach hooks in integration tests

adriancole:
We should only use beforeEach when state is required, but not known
until test execution (ex test name). Otherwise, it is distracting.

This removes unnecessary beforeEach and related lazy instantiation
from integration tests.


adriancole:
note: because time is dominated by docker and kafka tests, there's no measurable runtime speed increase due to this change.

adriancole:
on laptop, tests including browser and typescript finish in less than a minute. However, the kafkajs tests consume over half that time.

PR #430, opened 2019-07-27
Creates common test fixture for http server middleware

adriancole:
By using a common http server test fixture, we can remove needless
difference between instrumentation tests. This makes it easy to spot and
correct missing tests, and also easier to identify special behaviors.

Similar to #427, but for servers


jcchavezs:
It looks very promising.

lør. 27. jul. 2019, 09:05 skrev Adrian Cole <notifications@github.com>:

adriancole:
goal is no special tests.. hope to finish soon :)

adriancole:
finished the route-based name tests for express.

taking a break until tomorrow. There are some tests only hapi has I need to pull up

jcchavezs:
Curious about the route base span name test as I think it is only supported
by express and new instrumentations. Need a help to make it for all server
instrumentations?

lør. 27. jul. 2019, 14:20 skrev Adrian Cole <notifications@github.com>:

adriancole:
yeah it is best to support this anywhere it can be supported. one benefit
of tests is you can now notice things not supported :)

adriancole:
this is ready at this point. while it would be nice to have a meta-route builder to reduce the coupling between the server integration tests and their runners, it is too much more work to do at the moment. This change at least ensures all servers are running the same tests or it is visible that they aren't.

adriancole:
I managed to script route creation. This means we can add new tests without N changes unless the nature of routes change

PR #433, opened 2019-07-29
chore(#343): do not fail on 1xx responses.

### LINKED ISSUE #343: Status code 1xx should not be marked as error
Comes from https://github.com/openzipkin/brave/issues/874

Ping @adriancole @ghermeto

jcchavezs:
This PR disables failure on 1xx responses.

Ping @adriancole 

Closes #343

### DIFF
--- a/test/httpClientTestFixture.js
+++ b/test/httpClientTestFixture.js
@@ -141,7 +141,11 @@ function setupBasicHttpClientTests({clientFunction, requestScoped = false}) {
           'http.path': path,
           'http.status_code': '101',
         }
-      }));
+      }))
+      .catch((err) => {
+        /* eslint-disable no-console */
+        console.log(err);
+      });
   });
 
   it('should report 401 in tags', () => {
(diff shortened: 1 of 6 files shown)


adriancole:
please add a base test for this? it is not fun to do whack-a-mole changes later.. since on the brain, let's fix it for all

jcchavezs:
> please add a base test for this? it is not fun to do whack-a-mole changes later.. since on the brain, let's fix it for all

totally, I am still not done with the PR.

test/httpClientTestFixture.js
```diff
@@ -129,6 +129,25 @@ function setupBasicHttpClientTests({clientFunction, requestScoped = false}) {
...
+        name: 'get',
+        kind: 'CLIENT',
+        localEndpoint: {serviceName: localServiceName},
+        remoteEndpoint: {serviceName: remoteServiceName},
+        tags: {
+          'http.path': path,
+          'http.status_code': '101',
+        }
+      }))
+      .catch((err) => {
```
adriancole: odd you would need a catch here as the promise should bubble up. did you try the debug approach in the README? sometimes slowing down by using debugger can help faster vs guessing

test/middleware.js
```diff
@@ -2,6 +2,9 @@ const express = require('express');
 
 function middleware() {
   const api = express();
+  api.get('/weather/taipei', (req, res) => {
+    res.send(101);
```
adriancole: 101 is a clever number, but "switching protocols" will result in some strange behavior.

probably 100 Continue is best, but you will need a POST endpoint to handle this.

PR #436, opened 2019-08-09
Fixes lerna publish

jcchavezs:
This PR includes a step in the publish process that cleans up all old builds and avoids old code to show up in the `lib` folder of the node modules. This not only reduces the amount of files in the `lib` but also avoids confusions when debugging. Notice that we already used `lerna clean` but that only cleans the `node-modules` folder.

Important thing happens here: https://github.com/openzipkin/zipkin-js/pull/436/files#diff-cce1d632ab9e7d2789dc6ade2f34650cR8

Thanks a lot to @ca057

Ping @adriancole


package.json
```diff
@@ -33,15 +33,11 @@
     "@babel/preset-typescript": "^7.3.3",
     "@babel/register": "^7.3.3",
     "@types/chai": "^4.1.7",
-    "@types/mocha": "^5.2.5",
-
-    "express": "^4.17.0",
     "@types/express": "^4.17.0",
-
+    "@types/mocha": "^5.2.5",
```
jcchavezs: This is just a re-sorting. Nothing important.

packages/zipkin/rollup.config.js
```diff
@@ -5,6 +5,7 @@ import replace from 'rollup-plugin-replace';
 import {terser} from 'rollup-plugin-terser';
 import globals from 'rollup-plugin-node-globals';
 import builtins from 'rollup-plugin-node-builtins';
+import del from 'rollup-plugin-delete';
```
jcchavezs: ^ This is the plugin doing the trick for the clean up of old build files.

adriancole: cool

adriancole:
lgtm on green

PR #438, opened 2019-08-12
docs: improves batch reporter docs.

jcchavezs:
Ping @adriancole


packages/zipkin/src/batch-recorder.js
```diff
@@ -62,7 +62,7 @@ class BatchRecorder {
    * @constructor
    * @param {Object} options
    * @property {Logger} logger logs the data to zipkin server
-   * @property {number} timeout timeout for span in microseconds
+   * @property {number} timeout timeout for reporting spans in **microseconds**
```
adriancole: timeout after which an unfinished span is flushed to zipkin in microseconds

packages/zipkin/src/batch-recorder.js
```diff
@@ -82,7 +82,7 @@ class BatchRecorder {
           this._writeSpan(id, span);
         }
       });
-    }, 1000);
+    }, 1000); // runs every second
```
adriancole: // every second, this will flush to zipkin any spans that have timed out

jcchavezs:
@adriancole added more advices. PTAL

adriancole:
thanks @jcchavezs

PR #439, opened 2019-08-12
Added agent option to http transport

RAWeber:
Provide option to set a custom agent to use with node-fetch.

https://www.npmjs.com/package/node-fetch#custom-agent

This will allow users to specify any networking related options. For example, users can provide client certificates to hit mTLS endpoints, or CA certificates to use self-signed certificates.

Will use default agent if no agent option passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint:ts": "tslint -c tslint.json -e './packages/**/node_modules/**/*.ts' './packages/**/*.ts'",
     "lint": "npm-run-all --parallel lint:*",
     "test:es": "npm run lerna-test",
-    "test:ts": "tsr packages/zipkin/test-types/*.test.ts --noAnnotate --libDeclarations && mocha --opts test/mocha-types.opts",
+    "test:ts": "tsr packages/*/test-types/*.test.ts --noAnnotate --libDeclarations && mocha --opts test/mocha-types.opts",
     "test": "npm run test:es && npm run test:ts",
     "lerna-test": "lerna run test",
     "lerna-test-browser": "lerna run test-browser",
--- a/packages/zipkin-transport-http/index.d.ts
+++ b/packages/zipkin-transport-http/index.d.ts
@@ -1,6 +1,6 @@
-import {JsonEncoder, Logger, model} from 'zipkin';
 import {Agent as HttpAgent } from 'http';
 import {Agent as HttpsAgent} from 'https';
+import {JsonEncoder, Logger, model} from 'zipkin';
 
 type Agent = HttpAgent | HttpsAgent;
 
--- a/packages/zipkin-transport-http/index.d.ts
+++ b/packages/zipkin-transport-http/index.d.ts
@@ -1,4 +1,8 @@
 import {JsonEncoder, Logger, model} from 'zipkin';
+import {Agent as HttpAgent } from 'http';
+import {Agent as HttpsAgent} from 'https';
+
+type Agent = HttpAgent | HttpsAgent;
 
 declare class HttpLogger implements Logger {
   constructor(options: {
@@ -7,6 +11,7 @@ declare class HttpLogger implements Logger {
     jsonEncoder?: JsonEncoder,
     httpTimeout?: number,
     headers?: { [name: string]: any },
+    agent?: Agent | (() => Agent),
     log?: Console
   });
   logSpan(span: model.Span): void;
--- a/packages/zipkin-transport-http/index.d.ts
+++ b/packages/zipkin-transport-http/index.d.ts
@@ -9,11 +9,13 @@ declare class HttpLogger implements Logger {
     endpoint: string,
     httpInterval?: number,
     jsonEncoder?: JsonEncoder,
-    httpTimeout?: number,
+    timeout?: number,
+    maxPayloadSize?: number,
     headers?: { [name: string]: any },
     agent?: Agent | (() => Agent),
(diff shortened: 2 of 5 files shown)


adriancole:
let's default this to the current version of the transport. ex

`zipkin-transport-http/0.19.0`

While not here, we've had quite a discussion elsewhere on it. https://github.com/openzipkin/zipkin-reporter-java/issues/142

RAWeber:
Based on node-fetches [docs](https://nodejs.org/api/http.html#http_class_http_agent) and the nodejs http.Agent [docs](https://nodejs.org/api/http.html#http_class_http_agent) which it refers, I believe the user-agent referred to in: openzipkin/zipkin-reporter-java#142 would simply be a matter of updating the header[user-agent] property, while the Agent option I've added would refer to an http.Agent.

The http.agent would be responsible for managing connection persistence and reuse for HTTP clients, such as sockets, certs, etc.

I believe `zipkin-transport-http/0.19.0` could just be added to the default headers. I can add that to this commit if you would like, but it is not necessarily related to the agent parameter.

The current default uses the http.globalAgent, which has the base [defaults](https://nodejs.org/docs/latest-v7.x/api/http.html#http_new_agent_options) (this would reflect current behavior where no http.agent is being used)

I've also updated the `index.d.ts` file to reflect the proper type of the agent option.

adriancole:
sorry wrong agent I guess!

packages/zipkin-transport-http/index.d.ts
```diff
@@ -1,12 +1,17 @@
...
+type Agent = HttpAgent | HttpsAgent;
+
 declare class HttpLogger implements Logger {
   constructor(options: {
     endpoint: string,
     httpInterval?: number,
     jsonEncoder?: JsonEncoder,
     httpTimeout?: number,
     headers?: { [name: string]: any },
+    agent?: Agent | (() => Agent),
```
adriancole: is `(() => Agent)` the same effect as null?

Let's add typescript and javascript tests for this parameter to make sure someone later doesn't accidentally break what you need.

RAWeber: `(() => Agent)` is the equivalent of a function which returns an Agent

This follows node-fetch which will accept an HTTP(S) Agent or a function which returns a HTTP(S) Agent

package.json
```diff
@@ -14,7 +14,7 @@
     "lint:ts": "tslint -c tslint.json -e './packages/**/node_modules/**/*.ts' './packages/**/*.ts'",
     "lint": "npm-run-all --parallel lint:*",
     "test:es": "npm run lerna-test",
-    "test:ts": "tsr packages/zipkin/test-types/*.test.ts --noAnnotate --libDeclarations && mocha --opts test/mocha-types.opts",
+    "test:ts": "tsr packages/*/test-types/*.test.ts --noAnnotate --libDeclarations && mocha --opts test/mocha-types.opts",
```
RAWeber: Updated this to process `test-type/*.test.ts` tests in all packages. This reflects the second mocha command  which   uses `test/mocha-types.opts` and attempts to run test-type tests in all packages.

RAWeber:
Build failed on the Kafka Transport. The only change which could affect this (at least based on my current understanding of the project) is the change to the `test:ts` however, it failed on the `test:es` script, so I do not believe this was the cause.

Is there possibly another issue going on? Or an intermittent issue unrelated to these changes?

packages/zipkin-transport-http/test-types/HttpLogger.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      agent: new HttpAgent(),
+      log: console
+    };
+    const httpLogger: HttpLogger = new HttpLogger(options);
+
+    expect(httpLogger.logSpan).to.be.a('function');
+  });
+
+  it('should accept Http(s) Agent or function which returns Agent', () => {
+    const agents = [new HttpAgent(), new HttpsAgent(), () => new HttpAgent(), () => new HttpsAgent()];
```
adriancole: so this test makes sure it doesn't crash in the constructor. am I understanding that right?

I wonder if it would crash later. It might be confusing to have a test without assertions I mean.. wdyt?

adriancole:
kicked build in case it is a flake

packages/zipkin-transport-http/test-types/HttpLogger.test.ts
RAWeber: Ya it's testing the constructor. I've added an expect statement to confirm `this.agent` is properly being set and defaulting to `null` if it is not provided.

Beyond that, it is only being forwarded to the fetch call. Ideally I could confirm that fetch is being called with the agent, however due to the `node-fetch` implementation and `sinon` restrictions, it doesn't seem feasible to perform such a test without some hacky solutions.

jcchavezs:
Sorry I am too late for this discussion.

Out of curiosity: is this agent is also the same agent supported in axios:
https://github.com/axios/axios/blob/master/README.md#request-config

In addition, I feel like we haven't added any sort of documentation about
this new feature. Could we add at least in tsdoc about this new parameter?

In any case great work @RAWeber!


man. 19. aug. 2019, 14:30 skrev Adrian Cole <notifications@github.com>:

RAWeber:
@jcchavezs looks to be the same agent supported by Axios, as both just accept a Nodejs http(s) Agent.

And I will look into adding some documentation this week!

PR #440, opened 2019-08-14
Fix typo for zipkin-instrumentation-kafkajs

astik:
Fix required package name from the code example.


devinsba:
Awesome thanks! I'll let one of the JS folks merge but these types of fixes are greatly appreciated!

adriancole:
thanks

PR #442, opened 2019-08-22
Update README and JSDocs for http transport

RAWeber:
Reflects addition of http(s) agent, along with providing some extra information about existing options.


adriancole:
I don't have a link handy, but I can see some trouble around raising change off the master branch. You may be able to find some tips with a search like "github branch flow" or process.. basically this will prevent from having weird-looking commits in your change... the tl'dr being don't raise a pull request from your `master` branch

packages/zipkin-transport-http/index.d.ts
```diff
@@ -7,6 +7,18 @@ import {JsonEncoder, Logger, model} from 'zipkin';
...
+   * @constructor
+   * @param {Object} options
+   * @param {string} options.endpoint HTTP endpoint which spans will be sent
+   * @param {number} options.httpInterval How often to sync spans.
+   * @param {JsonEncoder} options.jsonEncoder JSON encoder to use when sending spans.
+   * @param {number} options.timeout Timeout for HTTP Post when sending spans.
+   * @param {number} options.maxPayloadSize Max payload size for zipkin span.
+   * @param {Object<string, any>} options.headers Any additional HTTP headers to be sent with span.
+   * @param {Agent|Function} options.agent HTTP(S) agent to use for any networking related options.
+   * @param {Console} options.log Internal logger used within the transport.
```
adriancole: is Console the right type here? I think above you suggested noop as an option

packages/zipkin-transport-http/index.d.ts
```diff
@@ -7,6 +7,18 @@ import {JsonEncoder, Logger, model} from 'zipkin';
...
 declare class HttpLogger implements Logger {
+  /**
+   * @constructor
+   * @param {Object} options
+   * @param {string} options.endpoint HTTP endpoint which spans will be sent
+   * @param {number} options.httpInterval How often to sync spans.
+   * @param {JsonEncoder} options.jsonEncoder JSON encoder to use when sending spans.
+   * @param {number} options.timeout Timeout for HTTP Post when sending spans.
+   * @param {number} options.maxPayloadSize Max payload size for zipkin span.
+   * @param {Object<string, any>} options.headers Any additional HTTP headers to be sent with span.
```
adriancole: guessing it would string, string unless we are being very lenient here

adriancole:
Going to merge this. any future polish can be done on our master prior to next release.

Thanks for all the help!

PR #446, opened 2019-09-17
zipkin: traceid should be <2**63

baloo:
scribe defines trace_id as i64:
      https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L431

To my understanding, we should not interpret the 64bit hex-encoded as i64 either, I think it's safer to just generate id in the `[0; 2**63[` range.

### DIFF
--- a/packages/zipkin/src/tracer/randomTraceId.js
+++ b/packages/zipkin/src/tracer/randomTraceId.js
@@ -2,10 +2,13 @@
 function randomTraceId() {
   const digits = '0123456789abcdef';
   let n = '';
-  for (let i = 0; i < 16; i += 1) {
+  for (let i = 0; i < 15; i += 1) {
     const rand = Math.floor(Math.random() * 16);
     n += digits[rand];
   }
+  const rand = Math.floor(Math.random() * 8);
+  n = digits[rand] + n;
+
   return n;
 }
 


adriancole:
Are you responding to a problem? We don't use numeric interpretation. Please prefer the swagger as that's current and valid (or bug) https://zipkin.io/zipkin-api/#/default/post_spans

baloo:
@adriancole I am sending that to a python implementation, which whenever it tried to serialize to thrift was throwing exception. I believe the same should happen with javascript although I haven't tried.

adriancole:
I think this change is at the wrong abstraction because the trace ID has no constraints on which bits are set. I'd look into whatever endpoint you are processing with.

For example, back when our server was based on thrift, we treated i64 just as a predefined array of 8 bytes. We didn't numerically interpret it.

jcchavezs:
I think it is easier that you tell us what the traceID is happening to fail
and also what is the library for python you are using?

José Carlos Chávez


tir. 17. sep. 2019 kl. 09:43 skrev Adrian Cole <notifications@github.com>:

baloo:
the python library is https://github.com/Gandi/python-zipkin/ which is only used by us. The error looks like:
```
  File "/usr/lib/python2.7/dist-packages/thriftpy2/thrift.py", line 153, in write
    oprot.write_struct(self)
  File "cybin.pyx", line 477, in cybin.TCyBinaryProtocol.write_struct (thriftpy2/protocol/cybin/cybin.c:6630)
  File "cybin.pyx", line 474, in cybin.TCyBinaryProtocol.write_struct (thriftpy2/protocol/cybin/cybin.c:6561)
  File "cybin.pyx", line 214, in cybin.write_struct (thriftpy2/protocol/cybin/cybin.c:3198)
thriftpy2.thrift.TDecodeException: Field 'trace_id(1)' of 'Span' needs type 'I64', but the value is `12759148914174905215L`
```

Trying to interpret traceid b1119a5629b4cb7f generated by zipkin-js

jcchavezs:
I don't think there is anything wrong with that `traceID`, I did a quick check (meaning I run the unit tests for that `traceID`) with `zipkin-php` and `zipkin-go` for example and everything was alright. I recommend you to use `py_zipkin` instead. On top of that, I am not sure about this assumption: https://github.com/Gandi/python-zipkin/blob/master/zipkin/util.py#L29 but @adriancole might have more information on this.

baloo:
@jcchavezs this comment is definitely wrong, and I am to fix it soon-ish. All I'm saying here is that there is little normalization on the traceid and considering the thrift definition file reads it should be an I64, sending a traceid that is `>2^63` sounds wrong. That's all I was intending to fix here.

PR #449, opened 2019-10-10
fix: fixes undefined comparison and partial span timeout rules.

jcchavezs:
This PR fixes a comparison with `undefined` and also adds a guard for `partial spans` that have not `timestamp` attribute (because they were already reported).

Ping @adriancole

### DIFF
--- a/packages/zipkin-instrumentation-express/src/expressMiddleware.js
+++ b/packages/zipkin-instrumentation-express/src/expressMiddleware.js
@@ -52,12 +52,25 @@ module.exports = function expressMiddleware({tracer, serviceName, port = 0}) {
 
       Object.defineProperty(req, '_trace_id', {configurable: false, get: () => id});
 
-      res.on('finish', () => tracer.letId(id, () => {
-        // if route is terminated on middleware req.route won't be available
-        const route = req.route && req.route.path;
-        tracer.recordRpc(instrumentation.spanNameFromRoute(req.method, route, res.statusCode));
-        instrumentation.recordResponse(id, res.statusCode);
-      }));
+      /**
+       * records response on finish or close (whichever happens first)
+       * @listens close
+       * @listens finish
+       */
+      const onCloseOrFinish = () => {
+        res.removeListener('close', onCloseOrFinish);
+        res.removeListener('finish', onCloseOrFinish);
+
+        tracer.letId(id, () => {
+          // if route is terminated on middleware req.route won't be available
+          const route = req.route && req.route.path;
+          tracer.recordRpc(instrumentation.spanNameFromRoute(req.method, route, res.statusCode));
+          instrumentation.recordResponse(id, res.statusCode);
+        });
+      };
+
+      res.once('close', onCloseOrFinish);
+      res.once('finish', onCloseOrFinish);
 
       next();
     }); // don't leak the trace ID from recordRequest
--- a/packages/zipkin/src/batch-recorder.js
+++ b/packages/zipkin/src/batch-recorder.js
@@ -129,7 +129,7 @@ class BatchRecorder {
     } else {
       isNew = true;
       // it can happen that timestamp is 0 hence this span will be always timed out
-      let timeoutTimestamp = (timestamp ? timestamp : now()) + this.timeout
+      const timeoutTimestamp = (timestamp || now()) + this.timeout;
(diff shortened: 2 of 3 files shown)


packages/zipkin/src/batch-recorder.js
```diff
@@ -105,8 +105,8 @@ class BatchRecorder {
 
   _writeSpan(id, span, isNew = false) {
     // TODO(adriancole) refactor so this responsibility isn't in writeSpan
-    if (!isNew && this.partialSpans.get(id) === 'undefined') {
-      // Span not found.  Could have been expired.
+    if (!isNew && this.partialSpans.get(id) === undefined) {
```
adriancole: my guess is this meant 'typeof' as undefined checks are a bit dodgy

```suggestion
    if (!isNew && typeof this.partialSpans.get(id) === 'undefined') {
```

packages/zipkin/src/batch-recorder.js
```diff
@@ -128,7 +128,9 @@ class BatchRecorder {
       span = this.partialSpans.get(id);
     } else {
       isNew = true;
-      span = new PartialSpan(id, timestamp + this.timeout);
+      // it can happen that timestamp is 0 hence this span will be always timed out
```
adriancole: interesting. any test case ideas?

packages/zipkin/src/batch-recorder.js
```diff
@@ -105,8 +111,8 @@ class BatchRecorder {
 
   _writeSpan(id, span, isNew = false) {
     // TODO(adriancole) refactor so this responsibility isn't in writeSpan
-    if (!isNew && this.partialSpans.get(id) === 'undefined') {
-      // Span not found.  Could have been expired.
+    if (!isNew && typeof (this.partialSpans.get(id)) === 'undefined') {
```
adriancole: not sure why the extra parens are here. will still need to test this method to ensure it isn't lying as it was before

packages/zipkin-instrumentation-express/src/expressMiddleware.js
```diff
@@ -52,12 +52,25 @@ module.exports = function expressMiddleware({tracer, serviceName, port = 0}) {
...
+       * @listens finish
+       */
+      const onCloseOrFinish = () => {
+        res.removeListener('close', onCloseOrFinish);
+        res.removeListener('finish', onCloseOrFinish);
+
+        tracer.letId(id, () => {
+          // if route is terminated on middleware req.route won't be available
+          const route = req.route && req.route.path;
+          tracer.recordRpc(instrumentation.spanNameFromRoute(req.method, route, res.statusCode));
```
jcchavezs: we are sure this is being reached, otherwise the integration test suite for httpserver will fail due to span naming.

packages/zipkin/src/batch-recorder.js
```diff
@@ -103,10 +109,10 @@ class BatchRecorder {
     span.delegate.setLocalEndpoint(span.localEndpoint);
   }
 
-  _writeSpan(id, span, isNew = false) {
+  _writeSpan(id, /** @type PartialSpan */ span, isNew = false) {
     // TODO(adriancole) refactor so this responsibility isn't in writeSpan
-    if (!isNew && this.partialSpans.get(id) === 'undefined') {
-      // Span not found.  Could have been expired.
+    if (!isNew && typeof this.partialSpans.get(id) === 'undefined') {
```
jcchavezs: In the end we did not need the parenthesis @adriancole

packages/zipkin/test/batch-recorder.test.js
```diff
@@ -486,4 +490,15 @@ describe('Batch Recorder - integration test', () => {
...
+
+  it('should not log span that is not in the reference', () => {
+    const span = {
+      delegate: {
+        timestamp: 1,
+        setLocalEndpoint: () => {}
+      }
+    };
+    recorder._writeSpan(rootId, span);
+    expect(spans).to.be.empty; // eslint-disable-line no-unused-expressions
```
jcchavezs: This test makes sure this is right: https://github.com/openzipkin/zipkin-js/pull/449/files#diff-503c850b99353bea68792242ff75f9edR114

packages/zipkin/src/batch-recorder.js
jcchavezs: Done.

PR #454, opened 2019-10-22
Call getLocalAddress() in HttpServerInstrumentation constructor

jdb8:
When profiling an express app using the zipkin express instrumentation, I noticed that for each request, around 15ms is being spent in `os.networkInterfaces()` . From reading the code, it looks like this only happens if no `host` is provided to the `HttpServerInstrumentation` class (if a host is provided when this class is instantiated, for each invocation of `recordRequest`, we don't need to look at the available network interfaces).

This is because the current fallback behaviour when no `host` is provided is to call [`InetAddress.getLocalAddress()`](https://github.com/jdb8/zipkin-js/blob/e65652100fd31ed41425e255f58b60e51a01602a/packages/zipkin/src/InetAddress.js#L34) each time [a new `LocalAddr` is created](https://github.com/jdb8/zipkin-js/blob/e65652100fd31ed41425e255f58b60e51a01602a/packages/zipkin/src/annotation.js#L65), which happens [every `recordRequest`](https://github.com/jdb8/zipkin-js/blob/master/packages/zipkin/src/instrumentation/httpServer.js#L86). By moving this call to the constructor, we can avoid doing it per-request and save around 15ms each time.

Example profile highlighting the time for `recordRequest` (18ms) before (you can see the long call to `networkInterfaces()`):

Screen Shot 2019-10-22 at 13 58 30

and a profile for the same server after this change (`recordRequest` at 3ms):

Screen Shot 2019-10-22 at 13 58 15

I'm assuming that this is a safe change as there's already support for providing a host explicitly, which remains in place, and the fallback behaviour is maintained albeit the IP address is no longer calculated on the fly per request. Hopefully this doesn't break any assumptions around the `host` value (I'll admit, I'm a little new to zipkin in general so please let me know if so).

Thanks!

### DIFF
--- a/packages/zipkin/src/InetAddress.js
+++ b/packages/zipkin/src/InetAddress.js
@@ -31,7 +31,7 @@ class InetAddress {
 }
 
 // In non-node environments we fallback to 127.0.0.1
-InetAddress.getLocalAddress = function getLocalAddress() {
+function getLocalAddress() {
   const isNode = typeof process === 'object' && typeof process.on === 'function';
   if (!isNode) {
     return new InetAddress('127.0.0.1');
@@ -40,6 +40,11 @@ InetAddress.getLocalAddress = function getLocalAddress() {
   // eslint-disable-next-line global-require
   const networkAddress = require('./network');
   return new InetAddress(networkAddress.ipv4());
-};
+}
+
+// Cache this value at import time so as to avoid network interface
+// lookup on every call
+const cachedLocalAddress = getLocalAddress();
+InetAddress.getLocalAddress = () => cachedLocalAddress;
 
 module.exports = InetAddress;
--- a/packages/zipkin/test/InetAddress.test.js
+++ b/packages/zipkin/test/InetAddress.test.js
@@ -5,6 +5,10 @@ describe('InetAddress', () => {
     InetAddress.getLocalAddress();
   });
 
+  it('should return the same object reference for multiple calls (cached)', () => {
+    expect(InetAddress.getLocalAddress()).to.equal(InetAddress.getLocalAddress());
+  });
+
   it('should convert an IP address to integer representation', () => {
     const addr = new InetAddress('80.91.37.133');
     expect(addr.toInt()).to.equal(1348150661);


adriancole:
hmm I'm wondering if there isn't a better way. the tracer itself should be providing the same (for all things, not just http server). let's look closer at where to put the fix.

(PS excellent snooping!)

adriancole:
yeah I think we should be handling this higher up, to solve more.

Right now, annotation.js is doing this, when we leave out the thing in httpServer or anywhere else.

```
  this.host = host || InetAddress.getLocalAddress();
```

the root problems are: InetAddress.getLocalAddress is not memoized. It should remember its value as no-where else (ex not in golang, java, ruby etc) do we expect to accommodate a changing local address. In other words, we can make this return a lazy-initialized constant (keeping the old signature)
```
// In non-node environments we fallback to 127.0.0.1
InetAddress.getLocalAddress = function getLocalAddress() {
```

Another problem is that we should deprecate the constructor for InetAddress. For example, if the address is localhost we definitely shouldn't attempt resolve etc.

```
class InetAddress {
  constructor(addr) {
    this.addr = addr;
  }
```

I presume that `getLocalAddress` will return the link local address, only falling back to localhost. So, when someone was formerly calling InetAddress(host), if it is "localhost" "127.0.0.1" or whatever I forget the ipv6 thing is, we return a constant, not  `getLocalAddress` 

So, deprecate `InetAddress(host)` for `InetAddress.get(host)` or whatever is idiomatic in javascript.

Mind giving a swing at this? You'll fix everywhere this way :)

PS how you ran profiler.. I don't really know how to do this, and I'm sure others don't either. If you have a tip to add to the README, much oblidged. You'll notice we recently added tips on how to use debugger. The more that know this, hopefully the less you worry about perf!

jdb8:
>the root problems are: InetAddress.getLocalAddress is not memoized. It should remember its value as no-where else (ex not in golang, java, ruby etc) do we expect to accommodate a changing local address. In other words, we can make this return a lazy-initialized constant (keeping the old signature)

Cool! Yeah, I had considered this but didn't want to make too big a change in case we did, for some reason, want to support that. Having it just memoize the value would be more generally useful.

Will try and find some time to work on it this week.

As for the profiling, I can add something to the README in my branch (or another, if you'd like): but essentially I span up an express server which uses the zipkin instrumentation, and used `--inspect` with node to attach a chrome debugger, then started recording a CPU profile and `curl`ed a basic endpoint (one that wasn't doing any real work other than processing tweens etc.). Basically following the same guide you have in your README, but more generally for a real server rather than the tests.

I noticed that there are integration tests for some or all of the instrumentation packages, but perhaps it'd also be nice to have an 'example.js' which spins up a minimal example and can be profiled? The code could be shared with the integration tests, etc. - just an idea for now.

adriancole:
> I noticed that there are integration tests for some or all of the
> instrumentation packages, but perhaps it'd also be nice to have an
> 'example.js' which spins up a minimal example and can be profiled? The code
> could be shared with the integration tests, etc. - just an idea for now.
>

would this be ok? in some other areas we use our example project also for
benchmarking https://github.com/openzipkin/zipkin-js-example/tree/master/web

jdb8:
>would this be ok? in some other areas we use our example project also for
benchmarking https://github.com/openzipkin/zipkin-js-example/tree/master/web

Yeah, that looks good! Thanks for the link.

I took another read of your comment before looking into expanding this PR. A couple of questions:

> the root problems are: InetAddress.getLocalAddress is not memoized. It should remember its value

This makes sense to me given your next sentence about not needing to handle a case where the IP changes. Given that `getLocalAddress` is the biggest slowdown, would memoizing this globally (say, at the module level) be enough? That would be pretty easy and require almost no code changes.

There's one small downside to memoizing vs. eagerly calling the function in the constructor, and that's that we'll still be incurring the slow (albeit, 15ms :) ) network-interface-lookup cost on the *first* `recordRequest`, rather than paying the cost at server initialisation time. However, if it makes the code simpler then it's likely acceptable since developers likely either don't care about such a small one-off cost or are warming their server sufficiently before accepting real requests that matter.

> Another problem is that we should deprecate the constructor for InetAddress. For example, if the address is localhost we definitely shouldn't attempt resolve etc.

I'm not sure I totally understand this part. From looking at the constructor, the whole class seems to just act as a box for easier coercion to and from different IP address formats.

>I presume that getLocalAddress will return the link local address, only falling back to localhost. So, when someone was formerly calling InetAddress(host), if it is "localhost" "127.0.0.1" or whatever I forget the ipv6 thing is, we return a constant, not getLocalAddress

I think I don't yet have enough familiarity with the code to spot this bug (assuming you're talking about a bug here): is this something that already exists or would be introduced if we added memoization?

If the memoization can be added independently of changing other APIs, I'd prefer to split this into two PRs (and dig in a bit more to understand the issue you're highlighting). If the two are more closely related, however, I'll need a bit more context to fully understand.

adriancole:
yep the changes are independent. Firstly memoize localhost. separately we can add an api such that is unlikely that something coming in as localhost, 127.0.01, etc independently does the same slow path that we just memoized.

adriancole:
also we should probably add docs that getLocalAddress is usually *not* localhost :P

link local is technically more likely I think, but renaming the function would probably break things.

jdb8:
> some sort of "is same as" assertion on the second call.

Like as a test? Or how do you mean?

Also, I've started having issues with the memcached instrumentation tests locally, getting errors like `Error: connect ECONNREFUSED 127.0.0.1:11211` - slightly worried these are related but ideally would check with CI: does travis run for PRs or only master?

jcchavezs:
My bad, I disabled CI for PRs on travis while working in canaries. I just reenabled them and they should work now.

adriancole:
yeah I meant in the test, if there's a way that for example getLocalAddress called twice is different, yet InetAddress.getLocalAddress (cached) is not.

jdb8:
Hmm, I guess I was treating the plain `getLocalAddress` as an implementation detail since it's not exported from the module (I could have inlined the logic rather than leaving it as a function at all, but it felt cleaner).

I can add a test that verifies that two calls to `InetAddress.getLocalAddress` return the same value? Although I would have expected that test to pass even on master, since the underlying IP shouldn't change during the test (hence the basis for why this memoization/caching is safe in the first place).

adriancole:
honestly I dont know if in javascript you can test difference between same
instance and same value. If not just skip.

thanks for thinking about it

jdb8:
Added a test, and verified that it fails on master.

jcchavezs:
This triggered a new canary release under version (0.19.2-alpha.1 in case
you want to test it.

fre. 25. okt. 2019, 03:39 skrev Adrian Cole <notifications@github.com>:

PR #464, opened 2020-01-24
chore(legal): adds MIT license

### LINKED ISSUE #463: Project has no OSS License

petermetz:
License itself could be anything else you choose of course. Submitting with MIT because this is the one I always use when given a choice.

Fixes #463

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 OpenZipkin
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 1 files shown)


jcchavezs:
Hi @petermetz thanks for this. I recommend we instead align with the other openzipkin libraries and use apache:
- https://github.com/openzipkin/brave/blob/master/LICENSE
- https://github.com/openzipkin/zipkin-go/blob/master/LICENSE

eirslett:
I guess somebody simply forgot to commit the license file. It's supposed to be Apache 2, like the other Zipkin repositories.

adriancole:
agreed. supposed to be ASL 2. no idea how we forgot so long :D

petermetz:
@eirslett @adriancole @jcchavezs 
Swapped it out to Apache 2.0, let me know if anything else is not cool. I just need it to have a legit license so I can use the project. ;-)


Not strictly related to this PR: While I was looking at the brave license file you linked I saw this (copyright info missing):
https://github.com/openzipkin/brave/blob/739a458dd5a20e8f0a57fb2b0a1d37998a038503/LICENSE#L189

jcchavezs:
Care to open a PR in brave?

man. 27. jan. 2020, 19:19 skrev Peter Somogyvari <notifications@github.com>:

jcchavezs:
Thanks a lot @petermetz

petermetz:
> Care to open a PR in brave? 

Sure: https://github.com/openzipkin/brave/pull/1068

adriancole:
ps ASL license text being literally copied has an advantage which is that it matches checksums with other projects and people don't have to do diffs on basis of the template. I don't think there's any reason for us to change the explanation section of the ASL text, we just need to apply the rules appropriate in our files which use the headers. ex this is one version of the ASL file which is indeed templated https://www.apache.org/licenses/LICENSE-2.0.txt

jcchavezs:
Good explanation @adriancole. Given that we are tackling this, do you think
we should use the same approach as brave and add license headers on the
files?

tir. 28. jan. 2020, 02:17 skrev Adrian Cole <notifications@github.com>:

adriancole:
possibly?

apache foundation has a short version they use in javascript files (to avoid assets being huge)

> Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements; and to You under the Apache License, Version 2.0.

https://www.apache.org/legal/src-headers.html#is-a-short-form-of-the-source-header-available

We could have something small like that if we desired, like..
> Copyright 2020 Copyright 2019 The OpenZipkin Authors; licensed You under the Apache License, Version 2.0.

I know that in our UI, we use the longer form headers, which we could also do here.

cc @openzipkin/core

PR #474, opened 2020-01-30
chore: Update dependencies to fix build failure.

anuraaga:
Found the magic to fix the build

- Add es6 lib explicitly
- Update typescript
- And had to adjust a flaky assertion


jcchavezs:
Thanks for this @anuraaga! I think in
https://github.com/openzipkin/zipkin-js/pull/474/files#diff-43e1f5ae0c8c03150bd1d6659c91a8c0R252
we need to update the path too.

While I think it is good to remove node 8, 9 (because EOL) I am still
curious about why it started failing, I mean what was the new versiom that
we don't explicitly handle that triggered this failure because most of our
dependencies are

Ping @sonybinhle <https://github.com/sonybinhle>

tor. 30. jan. 2020, 08:37 skrev Anuraag Agrawal <notifications@github.com>:

anuraaga:
Some dependencies specify `engines` in their package.json to just not allow the build on node 8. We can pass a flag to yarn to skip engine checks but if we're ok with removing the EOL builds from our travis I guess we don't need to.

anuraaga:
Restored Node 8 build by ignoring the engine checks. Since we've only bumped dev dependencies, I don't think there is any breaking change with this.

.travis.yml
```diff
@@ -17,45 +17,62 @@ matrix:
...
     - name: 'browser'
-      node_js: 10
+      node_js: 12
+      env:
+        - CXX=clang-3.6
+        - TEST_SUITE=browser
+      addons:
+        chrome: stable
+
+    - node_js: 13
```
jcchavezs: :+1:

jcchavezs:
Thanks a lot @anuraaga

PR #201, opened 2018-03-31
CLSContext now supports continuations with async-await

Ghost---Shadow:
`continuation-local-storage` has issues with `async-await` so someone made a library called [cls-hooked](https://www.npmjs.com/package/cls-hooked) which fixes that problem.

This PR merely changes the dependency from `cls` to `cls-hooked`.

Also, added a test case to demonstrate the phenomenon. 

Possible duplicate #135

Issue #175, opened 2017-12-21
Any plan for supporting async/await?


sleagon:
As mentioned in #124 and #135, this package does not work very well in node 8. Is there any schedule for something like zipkin-js@next or zipkin-js 2.x?

Fannon:
+1

adriancole:
some of the related issues are stuck on feedback. anyone want to comment on https://github.com/openzipkin/zipkin-js/pull/135#issuecomment-355870750 ?

Ghost---Shadow:
PR #201 adds this functionality

jcchavezs:
@joepuzzo would you be up for a PR fixing this (of course count us in for
reviews)

Den fre. 14. des. 2018, 16:24 skrev Joe Puzzo <notifications@github.com:

joepuzzo:
So i dont think you want me doing that tbh.. currently the lib uses continuation-local-storage which is the main supported lib.. the cls-hooks version is a fork that sounds like will eventually be merged into the main lib

joepuzzo:
The workaround is simple enough for now.. however one thing i could simply do is provide a second class AsyncCLSContext that was included with lib

jcchavezs:
Addressed by #476

Issue #106, opened 2017-05-29
Question: Promisified libraries


luismendozamx:
I've successfully integrated zipkin on some of our services. In most of them we are using Bluebird to [promisify](https://github.com/NodeRedis/node_redis#promises) the `redis.RedisClient` and we are using [`request-promise`](https://github.com/request/request-promise) for external service request, which also uses Bluebird internally.

What is the best approach to wrapping a promisified `redisClient` and `request-promise`? Should we use the `zipkin-instrumentation-redis` and `zipkin-instrumentation-request` packages and then promisify them? Or should we create our own instrumentation packages that support Promises?

eirslett:
I believe promisifying code shouldn't have any effect on zipkin instrumentation at all. Does it stop it from working?

luismendozamx:
I have only tried it for redis @eirslett.

From going through the instrumentation code I'm pretty sure why it doesn't work.

In the case of redis, using Bluebird to promisify it creates new methods with the `Async` word appended to [the method's name](https://github.com/NodeRedis/node_redis#promises). This commands are not taken into account in the ['redis-command' list](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-redis/src/zipkinClient.js#L2) used to wrap methods in the `zipkin-instrumentation-redis` package.

When I call an async redis method zipkin throws an error:
```
/app/my_app/node_modules/zipkin-instrumentation-redis/lib/zipkinClient.js:28
      callback.apply(this, args);
               ^

TypeError: callback.apply is not a function
    at zipkinCallback (/app/my_app/node_modules/zipkin-instrumentation-redis/lib/zipkinClient.js:28:16)
    at Object.callbackOrEmit [as callback_or_emit] (/app/my_app/node_modules/redis/lib/utils.js:89:9)
    at RedisClient.return_error (/app/my_app/node_modules/redis/index.js:701:11)
    at JavascriptRedisParser.returnError (/app/my_app/node_modules/redis/index.js:196:18)
    at JavascriptRedisParser.execute (/app/my_app/node_modules/redis-parser/lib/parser.js:560:12)
    at Socket.<anonymous> (/app/my_app/node_modules/redis/index.js:274:27)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
[...]

PR #476, opened 2020-01-31
feat: adds support for cls-hooked.

### LINKED ISSUE #124: Feature request: async_hooks based context

jcchavezs:
This PR adds support for `async` compatibility in CLS Context, using `cls-hooked` underneath and allowing users to opt-in this due to some [performance implications](https://github.com/DataDog/dd-trace-js/issues/695) for `cls-hooked` (update on readme needed).

**Why not a new package instead of adding it into existing package?**

In my mind `async` support is a feature on top of existing CLS package, it does not provide a fundamentally different functionality than it does already does but for modern versions of ES. In addition, it might become yet another package to support that is literally the same as CLS but with another library underneath.

**What about the performance implications in async_hooks**

This is already visible among node people and the situation [improved from node 8 to 12](https://github.com/DataDog/dd-trace-js/issues/695#issuecomment-538499129), it https://github.com/nodejs/node/issues/28420#issuecomment-506347295, in the other hand it is the only way to have async support using CLS, hence it feels like we have no choice.

Related to #201, #135

Closes #124,  #175, #106

Ping @javigomez @lussn @Ghost---Shadow @pbadenski @adriancole

### DIFF
--- a/packages/zipkin-context-cls/.babelrc
+++ b/packages/zipkin-context-cls/.babelrc
@@ -1,3 +1,4 @@
 {
-  "extends": "../../.babelrc"
+  "extends": "../../.babelrc",
+  "plugins": ["@babel/plugin-transform-runtime"]
 }
--- a/packages/zipkin-context-cls/README.md
+++ b/packages/zipkin-context-cls/README.md
@@ -24,7 +24,20 @@ the drawback then is that you have to pass around a context object manually.
 
 ## A note on CLS context and Promises
 
-This package is not suitable if your code inside the context uses promises. The context is then not properly propagated. There is work underway called [async_hooks](https://nodejs.org/api/async_hooks.html), but is at the time of this writing (node v10) in Experimental state.
+By default, this package is not suitable if your code inside the context uses promises, however you can enable an experimental feature for async/await support by using [cls_hooked](https://github.com/jeff-lewis/cls-hooked) library which underneath uses [async_hooks](https://nodejs.org/api/async_hooks.html).
+
+```javascript
+const CLSContext = require('zipkin-context-cls');
+const tracer = new Tracer({
+  ctxImpl: new CLSContext('zipkin', true),
+  recorder,
+  localServiceName: 'service-a'
+});
+```
+
+At the time of this writing, `async_hooks` [have some performance implications](https://github.com/DataDog/dd-trace-js/issues/695) whose effect may vary depending on the node version.
+
+The underneath implementation for async_hooks may change, that is why we hide that with the opt-in parameter.
 
 ## A note on the workings of CLS context
 
--- a/packages/zipkin-context-cls/package.json
+++ b/packages/zipkin-context-cls/package.json
@@ -13,6 +13,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "continuation-local-storage": "^3.1.7"
+    "continuation-local-storage": "^3.1.7",
+    "cls-hooked": "^4.2.2"
   }
 }
--- a/packages/zipkin-context-cls/package.json
+++ b/packages/zipkin-context-cls/package.json
@@ -13,10 +13,11 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "continuation-local-storage": "^3.1.7",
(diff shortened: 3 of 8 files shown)


eirslett:
Would it be reasonably to _only_ suppor cls-hooked? (Drop support for Node.js < 8.x)? Version 6.x had end-of-life April 30, 2019. Even version 8.x reached end of life a month ago.

jcchavezs:
Thanks for the feedback @eirslett 

> Would it be reasonably to only support cls-hooked?

Given that `cls-hooked` has some [performance issues](https://github.com/openzipkin/zipkin-js/pull/201#issuecomment-378436554) I'd rather keep both implementations, allowing the user to decide whether they want support async or not and inform about the impact. Otherwise we might penalize those who don't use async and are impacted. Notice I did abstracted the support for async and left `cls-hooked` as an implementation detail as I believe we might need to change this library in the future.

> (Drop support for Node.js < 8.x)? Version 6.x had end-of-life April 30, 2019. Even version 8.x reached end of life a month ago.

Currently we only support node 8, 10, 12 and 13. We would probably drop node 8 soon.

jcchavezs:
Ping @mziemer21 and @anujshankar who were also interested on this feature.

eirslett:
From the code, it looks like it's opt-out rather than opt-in, the default is supportAsync=false, which then uses cls-hooked? Maybe it could be named something different than `supportAsync` (what does that really mean?)
Anyways, great that support for this is finally being added!

javigomez:
cc: @scabfos

jcchavezs:
> From the code, it looks like it's opt-out rather than opt-in, the default is supportAsync=false, which then uses cls-hooked

Totally. I fixed that, thanks for the heads up!

jcchavezs:
Whoever wants to give it a try, can use `zipkin-context-cls@0.19.2-alpha.7` version.

eirslett:
Just a last comment from me; should the constructor take an options object maybe, instead of multiple arguments?

jcchavezs:
@eirslett that sounds good to me. I can change it.

jcchavezs:
This PR is pending to be merged until someone interested on the feature can try zipkin-context-cls@0.19.2-alpha.7.

kurtharriger:
I've written a unit test that reproduces the issue with current `CLSContext` implementation and verifies using `cls-hooked` fixes the issue.  

https://github.com/kurtharriger/zipkin-express-context-test

Ideally, perhaps this test should copied into the project itself, perhaps in `zipkin-instrumentation-express` project or something.

jcchavezs:
Hi @kurtharriger this is truly great. Would you open a PR vs this branch (`async_await`) with a failing and a successfull case (if possible) and I can merge and release this ASAP?

kurtharriger:
I've created 2 PRs.

The following adds the test I created above to the zipkin-instrumentation-express project.

https://github.com/openzipkin/zipkin-js/pull/498

I initially tried to update the CLSContext tests directly since I figure the behavior is not specific to express, but when I attempted to update the tests I found t didn't actually work without the express middleware as it seems event emitters need to be registered with the context implementation and once I did that I the test passed with either implementation.  So it seems the express specific test is still somewhat valuable as I'm not sure how to recreate that issue directly. 

https://github.com/openzipkin/zipkin-js/pull/499

mikhailrojo:
I added it to my project - and it works just fine
I enjoyed the idea - that you need only to put `true` to the context and it works out of the box

jcchavezs:
It is so great when a user can confirm this solves their problem.

Thanks everybody.

mikhailrojo:
@jcchavezs when I instrall zipkin-context-cls@0.21.0 I still get the old version without cls-hooked
has it been published?
can you please check why is that?

jcchavezs:
I will do a release tomorrow. I will see if I can get a couple of more improvements. E.g. https://github.com/openzipkin/zipkin-js/pull/505

PR #485, opened 2020-02-10
feat(zipkin-transport-http): configurable retry

petermetz:
Defaults to an exponential backoff with retrying always and forever.
All options are configurable through the constructor of the
HttpLogger class.

Goal: Do not lose traces if there was an intermittent network
connectivity issue in a browser environment for example.

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

### DIFF
--- a/packages/zipkin-transport-http/README.md
+++ b/packages/zipkin-transport-http/README.md
@@ -1,5 +1,5 @@
 
-# Zipkin-transport-http
+# zipkin-transport-http
 
 ![npm](https://img.shields.io/npm/dm/zipkin-transport-http.svg)
 
@@ -28,6 +28,7 @@ const recorder = new BatchRecorder({
     maxPayloadSize: 0 // Max payload size for zipkin span. Optional (defaults to 0)
     agent: new http.Agent({keepAlive: true}), // Agent used for network related options. Optional (defaults to null)
     log: noop // Logger to use. Optional (defaults to console)
+    fetchImplementation: require('node-fetch') // Pluggable fetch implementation (defaults to global fetch or fallsback to node fetch)
   })
 });
 
@@ -37,13 +38,19 @@ const tracer = new Tracer({
   localServiceName: 'service-a' // name of this application
 });
 ```
+
 ## Options
+
 ### Required
+
 - **endpoint** - HTTP endpoint which spans will be sent.
+
 ### Optional
+
 - **agent** - HTTP(S) agent to use for any networking related options.
   - Takes an [http](https://nodejs.org/api/http.html#http_class_http_agent)/[https](https://nodejs.org/api/https.html) NodeJS Agent.
   - Defaults to null
+
 ```javascript
 // Example using a self-signed CA, along with cert/key for mTLS.
 new HttpLogger({
@@ -55,6 +62,7 @@ new HttpLogger({
   })
 })
 ```
+
 - **headers** - Any additional HTTP headers to be sent with span.
   - Will set `'Content-Type':  'application/json'` at a minimum
 - **httpInterval** - How often to sync spans.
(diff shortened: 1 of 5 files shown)


jcchavezs:
Hi @petermetz thanks for this great addition. In terms of the code it looks good to me **tho** [we had a discussion about a similar issue](https://github.com/openzipkin/zipkin-go/pull/154#issuecomment-543422005) and we went for this path (suggested by @adriancole): 

> My main request is if there isn't a library that could be used that has resilience mechanics built-in, then users pass that library. this seems like commodity HTTP retry stuff which makes it feel odd we should host and maintain that.

Hence my question would be, does it make sense to use https://github.com/jonbern/fetch-retry instead of writing our own?

petermetz:
@jcchavezs Makes a lot of sense, thank you for the feedback! I swapped out the home grown retry logic with the library from your comment. 
The code looks much simpler and yet, it achieves the same thing. Let me know what you think.

petermetz:
@jcchavezs Also made it fully pluggable here: 027b5f8c80cb7aea1bdc32bc6325d61310b62acb

jcchavezs:
@petermetz thanks a lot for this great improvement. The only things that bothers me a lot is that HttpLogger accepts ad-hoc options for fetch-retry, I don't think it is its responsibility and also couples us with fetch-retry, also I think it becomes cumbersome as it gives the feeling that you could just pass the retry arguments and it will work with regular fetch. Hence I opened a PR in `fetch-retry` that allows to pass those options in the builder which will allows us to still accepting a fetch object: https://github.com/jonbern/fetch-retry/pull/39

petermetz:
@jcchavezs Got it, makes sense. Let's see how that PR plays out then, I can keep this open in the meantime and update this one accordingly based on what happens to https://github.com/jonbern/fetch-retry/pull/39

jcchavezs:
@petermetz I think now we can use
https://github.com/jonbern/fetch-retry/releases/tag/3.1.0. this will make
the code much simpler and I think it is a good idea to add some
documentation around the option of plugin a fetch client.

petermetz:
@jcchavezs Upgraded to 3.1.0 `fetch-retry`; simplified the constructor parameters as mentioned above and added an extra test + documentation with code examples on how to fully customize the fetch retry implementation:
- a6909d6f6a0f378cd4cc65862d2b38b1ad84bc45
- 0122eba23b3b900ec9080a318fd9ce9a3402dc1b
- 102fdd47693828377d03f09cd91a9c77d2da2648

jcchavezs:
Thanks for the great work @petermetz. I really appreciate your patience and enthusiasm.

I changed the PR to use the default `fetch` by default and added `fetch-retry` as an example. The main reason for this is that we should aim to provide the simpler/standard in the library but allow users to consciously extend the functionality of this library. Hence I kept the retry usage as an example.

jcchavezs:
José Carlos Chávez


---------- Forwarded message ---------
Fra: npm Inc <notifications@npmjs.com>
Date: man. 24. feb. 2020 kl. 12:02
Subject: Successfully published zipkin-transport-http@0.19.3-alpha.1
To: openzipkin <openzipkin-admin@googlegroups.com>


Hi openzipkin!

A new version of the package zipkin-transport-http (0.19.3-alpha.1) was
published at 2020-02-24T11:02:46.339Z from
34.66.200.49. The shasum of this package was
edf8df0c24dfc23decc6567c65a3314d6599016d.

If you have questions or security concerns, you can reply to this message or
email support@npmjs.com.

npm loves you.

-- 
You received this message because you are subscribed to the Google Groups
"OpenZipkin Admin" group.
To unsubscribe from this group and stop receiving emails from it, send an
email to openzipkin-admin+unsubscribe@googlegroups.com.
To view this discussion on the web visit
https://groups.google.com/d/msgid/openzipkin-admin/0101017076dd245c-baf0082e-b513-4836-b8ee-bcc23bc7cc70-000000%40us-west-2.amazonses.com
.

Issue #493, opened 2020-04-08
node_modules/zipkin-transport-http/src/HttpLogger.js.


smksnutmeg:
Running a Node application works fine for sending traces. However if I compile this to a minified file in Webpack, we lose the 'this' context after the interval when trying to do the fetch call for HttpLogger.

```
TypeError: this.fetchImplementation is not a function
    at HttpLogger.processQueue (/Users/shaun/nutmeg/nm-node-poc-web/build/api.bundle.js:51368:16)
    at Timeout._onTimeout (/Users/shaun/nutmeg/nm-node-poc-web/build/api.bundle.js:51301:13)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at unrefdHandle (timers.js:520:7)
    at Timer.processTimers (timers.js:222:12)
Waiting for the debugger to disconnect...
[nodemon] app crashed - waiting for file changes before starting...
```

```
    var timer = setInterval(function () {
      _this.processQueue();
    }, httpInterval);
```

 ```
 key: "processQueue",
    value: function processQueue() {
      var _this2 = this;

      var self = this;

      if (self.queue.length > 0) {
        var postBody = "[".concat(self.queue.join(','), "]");
        var fetchOptions = {
          method: 'POST',
          body: postBody,
          headers: self.headers,
          timeout: self.timeout,
          agent: self.agent
        };
        this.fetchImplementation(self.endpoint, fetchOptions).then(function (response) {
          if (response.status !== 202 && response.status !== 200) {
            var err = 'Unexpected response while sending Zipkin data, status:' + "".concat(response.status, ", body: ").concat(postBody);
            if (self.errorListenerSet) _this2.emit('error', new Error(err));else _this2.log.error(err);
          } else {
            _this2.emit('success', response);
          }
        })["catch"](function (error) {
          var err = "Error sending Zipkin data ".concat(error);
          if (self.errorListenerSet) _this2.emit('error', new Error(err));else _this2.log.error(err);
        });
        self.queue.length = 0;
        self.queueBytes = 0;
      }
    }
  }]);
```

**Comment by smksnutmeg** [comment]
Suggested solution
```

      const fetchImpl = this.fetchImplementation || defaultFetchImpl;

      fetchImpl(self.endpoint, fetchOptions).then((response) => {
```

jcchavezs:
Would you please come up with a PR for this?

jcchavezs:
[...]

Issue #491, opened 2020-03-25
Libraries as asap and node-fetch are used but they are not mentioned in deps


mikhailrojo:
Hi guys

I installed zipkin-instrumentation-request-promise, but got the error, that `asap` library is not installed. 

This asap library is mentioned in [devDeps section](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-request-promise/package.json#L23)
so it can not be used in production code as [import](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-instrumentation-request-promise/src/promise.js#L26)

the same for zipkin-transport-http [node-fetch is required](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-transport-http/src/HttpLogger.js#L6) but it is in [depsDev](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin-transport-http/package.json#L20)

O_o

jcchavezs:
Hi @mikhailrojo, thanks for opening this issue.

Regarding `asap` I think we indeed need to include it as a dependency in the package. Would you do that in a PR? cc @mkante 

The case of `node-fetch` is slightly different I think. I noticed this too when working on  https://github.com/openzipkin/zipkin-js/pull/485 but if you see `node-fetch` is a last resource as we first try to use other fetch implementations:

https://github.com/openzipkin/zipkin-js/blob/2aca37be1edca1b1e0bd62bcdb9ac20d1717c457/packages/zipkin-transport-http/src/HttpLogger.js#L2-L6

Can this be catched at compile time? If not, I think we make it clear in the readme, otherwise it sounds to me like even if there is a global fetch in place we are adding a new dependency to the user.

Any thoughts @adriancole @petermetz @anuraaga ?

anuraaga:
How about moving it to `peerDependencies`? From what I understand of the docs, it's how to specify an optional dependency like this one 

https://docs.npmjs.com/files/package.json#peerdependencies

jcchavezs:
Yeah I think `peerDependencies` is a good fit:

> NOTE: npm versions 1 and 2 will automatically install peerDependencies if they are not explicitly depended upon higher in the dependency tree. In the next major version of npm (npm@3), this will no longer be the case. You will receive a warning that the peerDependency is not installed instead. The behavior in npms 1 & 2 was frequently confusing and could easily put you into dependency hell, a situation that npm is designed to avoid as much as possible.

[...]

PR #497, opened 2020-05-02
feat(http): add fallback for fetch

rafael-anachoreta:
This PR adds a fallback value to fetch when `this` loses context.

I wasn't able to add any kind of tests to verify this change, but the existing ones still pass. 

Closes https://github.com/openzipkin/zipkin-js/issues/493

### DIFF
--- a/packages/zipkin-transport-http/src/HttpLogger.js
+++ b/packages/zipkin-transport-http/src/HttpLogger.js
@@ -58,7 +58,7 @@ class HttpLogger extends EventEmitter {
     this.timeout = timeout;
 
     const timer = setInterval(() => {
-      this.processQueue();
+      this.processQueue(this.fetchImplementation);
     }, httpInterval);
     if (timer.unref) { // unref might not be available in browsers
       timer.unref(); // Allows Node to terminate instead of blocking on timer
@@ -84,7 +84,7 @@ class HttpLogger extends EventEmitter {
   logSpan(span) {
     const encodedSpan = this.jsonEncoder.encode(span);
     if (this.maxPayloadSize && this._getPayloadSize(encodedSpan) > this.maxPayloadSize) {
-      this.processQueue();
+      this.processQueue(this.fetchImplementation);
       if (this._getPayloadSize(encodedSpan) > this.maxPayloadSize) {
         // Payload size is too large even with an empty queue, we can only drop
         const err = 'Zipkin span got dropped, reason: payload too large';
@@ -97,7 +97,7 @@ class HttpLogger extends EventEmitter {
     this.queueBytes += encodedSpan.length;
   }
 
-  processQueue() {
+  processQueue(fetchImpl) {
     const self = this;
     if (self.queue.length > 0) {
       const postBody = `[${self.queue.join(',')}]`;
@@ -108,8 +108,6 @@ class HttpLogger extends EventEmitter {
         timeout: self.timeout,
         agent: self.agent,
       };
-
-      const fetchImpl = this.fetchImplementation || defaultFetchImpl;
 
       fetchImpl(self.endpoint, fetchOptions).then((response) => {
         if (response.status !== 202 && response.status !== 200) {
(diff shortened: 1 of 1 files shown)


rafael-anachoreta:
Thanks for fixing the pipeline @jcchavezs 🙇 My local linting was all over the place, and I didn't realize travis had failed.

I can confirm that this change fixes the issue I was having with my webpack setup, is there anything else you'd like to me do before we can merge this in? Please advise!

smksnutmeg:
Hey guys - so you can replicate the issue if you compile this as a dependency with webpack. We bundle our node applications before sending to a docker image. I don't think this change will fix the issue.

jcchavezs:
@smksnutmeg is it possible for you to scratch a simple compilation step with webpack to reproduce the issue? cc @anuraaga @tacigar

smksnutmeg:
Yeah sure give me a bit of time.

smksnutmeg:
https://github.com/smksnutmeg/express-zipkin-example

jcchavezs:
Amazing @smksnutmeg, I wonder if we could integrate it into the https://github.com/openzipkin/zipkin-js-example (notice [we run the example every night](https://github.com/openzipkin/zipkin-js-example/blob/master/.circleci/config.yml#L32) against latest version of `zipkin-js` to make sure E2E works). This way we could avoid such issues in the future.

As far as I understand, when using webpack we can't access to the internal `this.fetchImpl` but I am not sure why `defaultFetchImpl` can't be accessible @smksnutmeg, is it maybe because of [this issue](https://github.com/openzipkin/zipkin-js/issues/491) and the error happening to you is that `node-fetch` is not under your deps (worth to hear your thoughts on that issue too)?

I see this wasn't an issue before https://github.com/openzipkin/zipkin-js/pull/485 (cc @petermetz) where we used a global `fetchImpl`. Since `processQueue` is an internal method, what do you think about mutating it into `processQueue(fetchImpl)`. My trade-off is that even if we rely on the `defaultFetchImpl` for the second execution (after interval), it becomes inconsistent as we give the user the idea that we accept it's `fetch` implementation but we only use it for the first time.

Finally I have may more questions now as I don't see how `this.fetchImplementation` does not exist but the attributes (`self.queue` or  `self.queueBytes`) do exist (cc @rafael-anachoreta).

rafael-anachoreta:
According to [this issue](https://github.com/webpack/webpack/issues/6379), it might actually be related to webpack/babel rather than to any problem with the actual code here. That said, it would still be amazing if zipkin could somehow handle that.

smksnutmeg:
Yeah looks like it could be. Definitely worth making backwards compatible nonetheless. Thanks!

jcchavezs:
@rafael-anachoreta or @smksnutmeg what do you think about what I proposed in :

> I see this wasn't an issue before #485 (cc petermetz) where we used a global fetchImpl. Since processQueue is an internal method, what do you think about mutating it into processQueue(fetchImpl). My trade-off is that even if we rely on the defaultFetchImpl for the second execution (after interval), it becomes inconsistent as we give the user the idea that we accept it's fetch implementation but we only use it for the first time.

Would that do the trick? Would you try it in a PR?

rafael-anachoreta:
Hey @jcchavezs I've pushed an update to the branch with what I understood as the changes you proposed. I have tried them locally on my webpack setup and it seemed to work just fine, but I'm not sure if that's exactly what you were looking for.

Please let me know!

jcchavezs:
Thanks @rafael-anachoreta ! I will add some comments to the code explaining the change after merge! Is there any change you or @smksnutmeg can reproduce the webpack inclusion in https://github.com/smksnutmeg/express-zipkin-example into https://github.com/openzipkin/zipkin-js-example so we make sure such things don't happen again?

jcchavezs:
@rafael-anachoreta you can eagerly try zipkin-transport-http@0.21.1-alpha.1 while I cut a release.

petermetz:
@jcchavezs 

> I see this wasn't an issue before #485 (cc @petermetz) where we used a global `fetchImpl`. Since `processQueue` is an internal method, what do you think about mutating it into `processQueue(fetchImpl)`. My trade-off is that even if we rely on the `defaultFetchImpl` for the second execution (after interval), it becomes inconsistent as we give the user the idea that we accept it's `fetch` implementation but we only use it for the first time.

Later for the party but I also think it's a good idea to reduce usage of globals as much as possible.

PR #498, opened 2020-05-02
tests: adds test to verify context during concurrent request

kurtharriger:
Adds a test for concurrent request that ensures that content is correctly preserved over promises when using cls-hooked. 

If you change 
```
ctxImpl = new CLSContext('zipkin', true); 
```
to 
```
ctxImpl = new CLSContext('zipkin', false);
```
the test will fail.


jcchavezs:
Thanks a lot @kurtharriger, could you please also add the failing case?

jcchavezs:
I'll add the test case myself. Thanks a lot @kurtharriger!

PR #499, opened 2020-05-02
Expose bindEmitter on CLSContext

kurtharriger:
I noticed that the existing test `supports async-await contexts` passes even if you disable the async await support on line 98.

I then tried to update this using an EventEmitter as I did in my other PR to more reliably reproduce the concurrent request behavior and was a bit surprised to discover that even with cls-hooked the test failed.   

It appears that when using an EventEmitter you need to explicitly register it with cls-hooked and when I did that it passed.  I then discovered that cls-continuation-context also supported a bindEmitter method and after exposing bindEmitter on context the test would pass in either implementation.  

I'm a bit sad it was necessary to explicitly register the emitter as I had hoped that when using async-hooks anything created within the context of an active namespace would automatically propagate the context, but it appears not to be the case.

So given that it is necessary to register EventEmitters with context this PR exposes that method and adds the test and I removed the existing test since it apparently does not fail when it should.

### DIFF
--- a/packages/zipkin-context-cls/test/CLSContext.test.js
+++ b/packages/zipkin-context-cls/test/CLSContext.test.js
@@ -1,3 +1,4 @@
+const EventEmitter = require('events');
 const CLSContext = require('../');
 
 const namespace = 'mynamespace';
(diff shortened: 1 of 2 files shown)


jcchavezs:
Thanks @kurtharriger. This is such a great work. I have two concerns here:
1. Can't we make `bindEmitter` private or use it for tests but skip them on compilation?
2. If not, I'd probably prefix it with an underscore or add comments saying it is for testing purposes only (cc @adriancole)

Other than that, truly gem of work!

packages/zipkin-context-cls/test/CLSContext.test.js
```diff
@@ -84,6 +85,59 @@ function CLSContextPerAsync(supportAsync) {
...
     });
   });
+
+  it('supports callbacks on eventEmitter', async() => {
+    const eventEmitter = new EventEmitter();
+    const ctx = new CLSContext(namespace, supportAsync);
+    // if you comment this out the tests will fail
+    // Apparently, it is necessary to explicitly register event emitters with context
+    ctx.bindEmitter(eventEmitter);
+    // const dbg = (...args) => console.log(...args);
```
jcchavezs, line 95: I actually think we need loggers in this kind of tests. What if we make the logger enabled or disabled based on `logLevel`? see https://karma-runner.github.io/4.0/config/configuration-file.html

jcchavezs:
Thanks a lot @kurtharriger !

PR #504, opened 2020-05-25
Context setup fix

mikhailrojo:
Context should be set up after callable run.
On the first run when it is null, otherwise we nullify it after callable sets it.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "lerna": "3.16.5",
-  "version": "0.21.0",
+  "version": "0.21.1",
   "command": {
     "run": {
       "npmClient": "yarn"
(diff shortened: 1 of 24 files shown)


jcchavezs:
Hi @mikhailrojo, thanks for opening this. I think I understand the change but I would feel more comfortable if we could include failing test for this. Would you be up for that?

mikhailrojo:
After seeing the tests - I realized this is intended behavour

my only surprize  is that it shouldn't work well with zipkin-instumentation-express 
as there we do `tracer.scoped(() => { setId});` and after this funcation is done.
the context id is set back to null.

And next function which calls tracer.getContext() -> gets null.
so may be a pull request should be directed to `zipkin-instumentation-express`

I am confused

jcchavezs:
@mikhailrojo do you think this is still an issue?

mikhailrojo:
well, the behaviour still seems strange to me... the functionality seems to me which can be used.
But as you merge cls-hooked - I am using that - and I am not going to use explicit context

PR #505, opened 2020-05-26
Deps correctness

mikhailrojo:
I moved libraries which are required in code in deps from devDeps and add commas to readme.
Closes https://github.com/openzipkin/zipkin-js/issues/491

### DIFF
--- a/packages/zipkin-instrumentation-request-promise/package.json
+++ b/packages/zipkin-instrumentation-request-promise/package.json
@@ -16,16 +16,16 @@
     "@types/request": "^2.0.8"
   },
   "dependencies": {
+    "asap": "^2.0.6",
+    "request": "^2.88.2",
+    "request-promise": "^4.2.5",
     "zipkin": "^0.21.0"
   },
   "devDependencies": {
     "@types/request": "^2.0.8",
-    "asap": "^2.0.6",
     "co": "^4.6.0",
     "co-mocha": "^1.2.1",
     "nock": "^9.0.14",
-    "request": "^2.83.0",
-    "request-promise": "^4.2.2",
     "uuid": "^3.2.1",
     "zipkin-context-cls": "^0.21.0"
   }
--- a/packages/zipkin-instrumentation-request-promise/package.json
+++ b/packages/zipkin-instrumentation-request-promise/package.json
@@ -16,7 +16,7 @@
     "@types/request": "^2.0.8"
   },
   "dependencies": {
-    "asap": "^2.0.6",
+    "asap": "^2.0.4",
     "request": "^2.88.2",
     "request-promise": "^4.2.5",
     "zipkin": "^0.21.0"
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -14,8 +14,10 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
-    "node-fetch": "^2.6.0",
     "zipkin": "^0.21.0"
+  },
+  "peerDependencies": {
+    "node-fetch": ">=1.4.0 <3.0.0"
   },
   "devDependencies": {
     "body-parser": "^1.15.2",
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -19,7 +19,6 @@
   },
   "devDependencies": {
     "body-parser": "^1.15.2",
-    "express": "^4.17.1",
     "fetch-retry": "3.1.0",
     "noop-logger": "^0.1.1"
   }
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -14,12 +14,13 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
+    "node-fetch": "^2.6.0",
     "zipkin": "^0.21.0"
   },
   "devDependencies": {
     "body-parser": "^1.15.2",
+    "express": "^4.17.1",
     "fetch-retry": "3.1.0",
-    "node-fetch": "^2.3.0",
     "noop-logger": "^0.1.1"
   }
 }
--- a/packages/zipkin-transport-http/README.md
+++ b/packages/zipkin-transport-http/README.md
@@ -25,9 +25,9 @@ const recorder = new BatchRecorder({
     httpInterval: 1000, // How often to sync spans. Optional (defaults to 1000)
     headers: {'Authorization': 'secret'}, // Custom HTTP headers. Optional
     timeout: 1000, // Timeout for HTTP Post. Optional (defaults to 0)
-    maxPayloadSize: 0 // Max payload size for zipkin span. Optional (defaults to 0)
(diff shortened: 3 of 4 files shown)


packages/zipkin-transport-http/package.json
```diff
@@ -14,12 +14,13 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
+    "node-fetch": "^2.6.0",
     "zipkin": "^0.21.0"
   },
   "devDependencies": {
     "body-parser": "^1.15.2",
+    "express": "^4.17.1",
```
jcchavezs, line 22: I am curious about this one, how come tests don't complain?

packages/zipkin-instrumentation-request-promise/package.json
```diff
@@ -16,16 +16,16 @@
     "@types/request": "^2.0.8"
   },
   "dependencies": {
+    "asap": "^2.0.6",
```
jcchavezs, line 19: :+1:

packages/zipkin-transport-http/package.json
mikhailrojo: My guess is as it is lerna, expressjs was coming from the global package.json

packages/zipkin-transport-http/package.json
```diff
@@ -14,12 +14,13 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/openzipkin/zipkin-js",
   "dependencies": {
+    "node-fetch": "^2.6.0",
```
jcchavezs, line 17: I am kind of concerned about this version constraint. I checked node-fetch api and nothing stop us to use [version 1.4.0](https://github.com/node-fetch/node-fetch/tree/v1.4.0) for example since it support all the options we use. In the other hand, [version 3.x is going to break us](https://github.com/node-fetch/node-fetch/commit/94e5b92de1d7bb3506ac6487581963a327e00d22#diff-2936d6bde71136a082ca21cd43f997e9R26). I feel like we could support ^1.4 and ^2.0? What do you think. I don't want us to force users to upgrade their node-fetch because of zipkin.

packages/zipkin-instrumentation-request-promise/package.json
```diff
@@ -16,16 +16,16 @@
...
+    "request": "^2.88.2",
+    "request-promise": "^4.2.5",
```
jcchavezs, line 21: I am not sure about this one. It feels like CLS is not supported in v4: https://github.com/request/request-promise#migration-from-v3-to-v4

Do you have any idea?

jcchavezs:
Thanks for this. I added some comments.

I know you are just moving devDeps into deps but that means we will impact user's code and hence it is worth taking the time to look at the deps we are aiming to introduce to our users.

packages/zipkin-instrumentation-request-promise/package.json
mikhailrojo: Jose, I set the version 03c84c680240a9210cf08e2a7c94a96d874df5db
but what is the point of fixing patch version?

packages/zipkin-transport-http/package.json
mikhailrojo: deleted express as it is in global package json

packages/zipkin-transport-http/package.json
mikhailrojo: why do you think this package uses 1.4.0 ? I moved this package from version 2.3 to 2.6 
you suggest to fix it here as ^2.0.0?

packages/zipkin-instrumentation-request-promise/package.json
mikhailrojo: I moved it from devDeps v 4.2.2 to deps 4.2.5 
where is v3?

packages/zipkin-transport-http/package.json
petermetz: > why do you think this package uses 1.4.0 ? I moved this package from version 2.3 to 2.6
> you suggest to fix it here as ^2.0.0?

I think you could specify it as `>=2.0.0 < 3.0.0` (could also be that simply `2.x` expresses the same thing but don't quote me on that, I've slipped up with semver constraints before)

Could also be `>=1.4.0 < 3.0.0` to include the 1.4.0 version as @jcchavezs mentioned

mikhailrojo: Did I get it right that ^2.0.0 will be fine for everyone?

petermetz: > Did I get it right that ^2.0.0 will be fine for everyone?

Would be fine by me personally because I don't use old node-fetch (`v1.4.x`) at all.

jcchavezs: Yeah, my main concern are those using 1.4.x that we will upgrade them to 2.0.x but now I think that since this is not a breaking change because we are just adding the dependency it is fine to keep ^2.0.0. Any thoughts @adriancole?

packages/zipkin-instrumentation-request-promise/package.json
jcchavezs: Sorry I meant `^2.0.4` but it is a patch so probably this is good enough :+1:

packages/zipkin-instrumentation-request-promise/package.json
jcchavezs: Yeah, right. Ignore this one.

jcchavezs:
Once this one https://github.com/openzipkin/zipkin-js/pull/505/files#r432086456 is sorted we can merge it and do a release.

packages/zipkin-transport-http/package.json
petermetz: > Yeah, my main concern are those using 1.4.x already that want to use zipkin-js, will that force them to use node-fetch 2.x then? 
>In the other hand I think that since this is not a breaking change because we are just adding the dependency it is fine to keep ^2.0.0. Any thoughts @adriancole?

If they want to upgrade to the newer version of zipkin-js, then yes, they will be forced to update node-fetch from 1.4.x to at least 2.x.
So, in this sense, this is a breaking change and should be accompanied by a major version bump on the version of zipkin-js itself.

Basically it's like this: If you previously did not restrict a peer dependency version, but now you do, then it is a breaking change for those who are suddenly out of compliance with their dependency version.

If it were released as a minor/patch version bump then it might break builds for some, if they had minor/patch version auto upgrade specified in their zipkin-js version in the package.json dependencies section. This is how builds of projects with npm dependency management can break overnight seemingly all by themselves without anyone changing anything in the source code.

So, I'd say it needs a major release.
That way people who are upgrading their zipkin-js will know to "expect trouble" and those who specified to only auto-update patch/minor versions won't get a sudden and unexpected breakage.

People who specified to auto-upgrade even major versions and use node-fetch 1.4.x will still have their build break overnight, but they chose to live on the bleeding edge so there's nothing you can do about that.

jcchavezs: Thanks for the explanation @petermetz, if that is the case then let's do `>=1.4.0 < 3.0.0` so we don't break anyone. We are not likely doing a major release for this change.

mikhailrojo: 3448533b8a01daf27da88f7e1b3e9fc9e8a7b62d

jcchavezs:
Thanks a lot @mikhailrojo !

PR #506, opened 2020-05-27
Lerna link convert

mikhailrojo:
see [lerna doc about common deps](https://github.com/lerna/lerna#common-devdependencies)

I just make `lerna link convert` - so all dependencies got hoisted to the root package.json
It will release  cluttered individual packages from the common devDeps


packages/zipkin-encoder-thrift/package.json
```diff
@@ -22,6 +22,6 @@
   "homepage": "https://github.com/openzipkin/zipkin-js#readme",
   "dependencies": {
     "thrift": "^0.11.0",
-    "zipkin": "^0.21.0"
+    "zipkin": "file:../zipkin"
```
jcchavezs, line 25: I see this link https://github.com/lerna/lerna#common-devdependencies specifically talks about `devDependencies` but this change changes `dependencies`, I am curious about what would be the version being pinned for this.

mikhailrojo: the current version in packages/zipkin
currently 0.21.0, but when upgraded 0.22.0

as lerna here works as in fixed(not independent mode) every time we change in one lib, every other upgrades automatically.

mikhailrojo:
I just realized that dependencies are managed by yarn workspaces - and this PR is useless
stupid me

jcchavezs:
We still appreciate the effort. Thanks a lot!

PR #510, opened 2020-08-05
add propagation interface

asanluis:
TODO if is the correct approach
 - Delete RequestZipkinHeader
 - Refactor kafkajs and expressjs instrumentation
 - Update README, add example usage of a custom propagation
 - Add AWS propagation module


packages/zipkin/src/tracer/index.js
```diff
@@ -37,13 +39,15 @@ class Tracer {
...
-    defaultTags
+    defaultTags,
+    propagation = new B3Propagation()
   }) {
     this.log = log;
     this.recorder = recorder;
     this.sampler = sampler;
     this.traceId128Bit = traceId128Bit;
     this.supportsJoin = supportsJoin;
+    this._propagation = propagation;
```
asanluis, line 50: Don't know if a better approach is delete the extract and injector methods and only add a get propagation, this simplify the grpc-client implementation, thats need to know the propagation headers....

packages/zipkin-instrumentation-express/src/wrapExpressHttpProxy.js
```diff
@@ -18,7 +18,7 @@ class ExpressHttpProxyInstrumentation {
       const clientTraceId = this.tracer.createChildId();
       this.tracer.setId(clientTraceId);
 
-      const proxyReqWithZipkinHeaders = Request.addZipkinHeaders(proxyReq, clientTraceId);
+      const proxyReqWithZipkinHeaders = this.tracer.injector(proxyReq);
```
jcchavezs, line 21: I'd rather pass the clientTraceId explicitly instead of having the injector to know about the scope.

packages/zipkin/index.d.ts
```diff
@@ -61,6 +62,11 @@ declare namespace zipkin {
     recordLocalAddr(inetAddress: InetAddress): void;
     recordBinary(key: string, value: boolean | string | number): void;
     writeIdToConsole(message: any): void;
+    /** Extract propagation ctx from request */
+    extractId(readHeader: <T> (header: string) => option.IOption<T>): void;
```
jcchavezs, line 66: I think I'd prefer to have the extractId to be actually an extractor like the injector.

packages/zipkin/index.d.ts
```diff
@@ -61,6 +62,11 @@ declare namespace zipkin {
...
+    /** Injector propagation ctx from request */
+    injector(request: any): object;
```
jcchavezs, line 68: injector can be for messaging too so I think it is better that we pass something like `readHeader` and call it `setter` and this return an object that performs the injection, does that make sense?

packages/zipkin/index.d.ts
```diff
@@ -368,6 +367,16 @@ declare namespace zipkin {
       recordError(traceId: TraceId, error: Error): void;
     }
   }
+  namespace propagation {
+      interface Propagation {
+        extractor<T>(tracer: Tracer, readHeader: <T> (header: string) => option.IOption<T>): TraceId;
```
jcchavezs, line 372: If I understand correctly, we pass the tracer to the extractor?

packages/zipkin/src/index.ts
```diff
@@ -16,7 +15,7 @@ export { default as ConsoleRecorder } from './console-recorder';
 export { default as ExplicitContext } from './explicit-context';
 
 export { default as Instrumentation } from './instrumentation';
-export { default as Request } from './request';
+export { default as B3Propagation } from './propagation/b3propagation';
```
jcchavezs, line 18: <3

packages/zipkin/src/instrumentation/httpServer.js
```diff
@@ -74,7 +27,7 @@ class HttpServerInstrumentation {
   }
 
   recordRequest(method, requestUrl, readHeader) {
-    this._createIdFromHeaders(readHeader).ifPresent(id => this.tracer.setId(id));
+    this.tracer.extractId(readHeader);
```
jcchavezs, line 30: I am really happy this becomes so simple now.

packages/zipkin/src/propagation/b3propagation.js
```diff
@@ -0,0 +1,75 @@
...
+    headers[this.headers.SpanId] = traceId.spanId;
+
+    traceId.parentSpanId.ifPresent((psid) => {
+      headers[this.headers.ParentSpanId] = psid;
+    });
+    traceId.sampled.ifPresent((sampled) => {
+      headers[this.headers.Sampled] = sampled ? '1' : '0';
+    });
+
+    if (traceId.isDebug()) {
```
jcchavezs, line 68: In here, if `traceId.isDebug()) then you don't need to pass sampling.

packages/zipkin/src/tracer/index.js
jcchavezs: I would go for that instead.

packages/zipkin/index.d.ts
jcchavezs: In new tracing models we have propagation decoupled from tracer, I think we should follow similar pattern here.

jcchavezs:
This is great work @asanluis I left you some comments cc @adriancole

packages/zipkin/index.d.ts
adriancole: agree we should not couple these

jcchavezs:
I will come back to it this week.

PR #512, opened 2020-08-23
Changed cls-context dependency continuation-locatl-storage for cls-hooked

zardilior:
Changed cls-context dependency continuation-locatl-storage for cls-hooked, due to the fact that cls-hooked uses new nodejs features that make cls work better and have less issues

### DIFF
--- a/packages/zipkin-context-cls/README.md
+++ b/packages/zipkin-context-cls/README.md
@@ -2,7 +2,7 @@
 
 ![npm](https://img.shields.io/npm/dm/zipkin-context-cls.svg)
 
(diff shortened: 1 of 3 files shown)


jcchavezs:
@zardilior thanks for coming by, could you please elaborate more on the "work better and have less issues" and more importantly, what node versions we aiim to? Check this comment to understand the concerns here: https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-context-cls#a-note-on-cls-context-and-promises

zardilior:
Just checked it out and it was already done, there is a flag for supporting await which then uses cls-hooked instead of the current library, closing this pr. Wasn't done some months ago when I made my local changes.

PR #517, opened 2020-10-08
[Proof of Concept] Broken isOptional Check

whizzzkid:
# Do Not Merge


jcchavezs:
@whizzzkid is this still valid?

whizzzkid:
We ended up upgrading all of our packages to the same semver to fix this issue.

PR #519, opened 2020-11-12
docs: adds note about defaultTags

jcchavezs:
Ping @adriancole @ghermeto

### DIFF
--- a/packages/zipkin/src/tracer/index.js
+++ b/packages/zipkin/src/tracer/index.js
@@ -39,7 +39,7 @@ class Tracer {
     localEndpoint,
     /* eslint-disable no-console */
     log = console,
-    // defaultTags allows the user to pass tahs that should be added to all spans by
+    // defaultTags custom tags to add to all sampled spans. Defaults to empty.
     // default.
     defaultTags
   }) {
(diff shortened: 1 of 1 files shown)


packages/zipkin/src/tracer/index.js
```diff
@@ -39,6 +39,8 @@ class Tracer {
     localEndpoint,
     /* eslint-disable no-console */
     log = console,
+    // defaultTags allows the user to pass tahs that should be added to all spans by
```
adriancole, line 42: I'd tighten it up though curious how this question came up.. where did they think the tags would go? maybe it can help revise the comment.

"defaultTags custom tags to add to all sampled spans. defaults to empty"

jcchavezs: I like documenting my issues so I was grabbing references to open https://github.com/hypertrace/agent-config/issues/21 then I realised some params have comments and some don't and hence I felt the need to improve this.

jcchavezs, line 42:
```suggestion
    // defaultTags custom tags to add to all sampled spans. Defaults to empty.
```

packages/zipkin/src/tracer/index.js
```diff
@@ -39,6 +39,8 @@ class Tracer {
     localEndpoint,
     /* eslint-disable no-console */
     log = console,
+    // defaultTags custom tags to add to all sampled spans. Defaults to empty.
+    // default.
```
codefromthecrypt, line 43:
```suggestion
```

codefromthecrypt (approved):
thanks for the help

Issue #465, opened 2020-01-27
Node standard "os" error


santiagoalmeidabolannos:
I am getting this `The package at "node_modules\zipkin\lib\index.js" attempted to import the Node standard library module "os".` it should not be happening.

I add this into my `tsconfig.json` 

`"paths": {
      ...
      "node-fetch": [
        "node_modules/empty-module/index.js"
      ],
      "os": [
        "node_modules/empty-module/index.js"
      ]
    },`

jcchavezs:
Could be related to https://github.com/openzipkin/zipkin-js/commit/b3b5a0ca7466f76e9139f2f9409d5af4f42ea3f7? cc @jdb8

santiagoalmeidabolannos:
exactly, I remove those lines and it works

jdb8:
Which version of node is producing this error? Can you paste the full traceback if there's more? I can't tell what's actually throwing the error here.

santiagoalmeidabolannos:
It is simple, I am running this on React, no Node standard library runs on it but base on the package description it shouldn't happen

jdb8:
Oh, this is running in a browser environment? I wonder if we need to wrap the import-level execution in a check for that?

santiagoalmeidabolannos:
doesn't look bad. Maybe this is helpful https://github.com/flexdinesh/browser-or-node#readme, without any extra package this can be used:
`var isBrowser=new Function("try {return this===window;}catch(e){ return false;}");`
or
`var isNode=new Function("try {return this===global;}catch(e){return false;}");`

jdb8:
@santiagoalmeidabolannos just for additional clarification, do you see this error in the browser console at runtime or in a warning at build-time? Just curious for future reference, and also to understand whether this breaks the build or causes errors or something else.

santiagoalmeidabolannos:
@jdb8 I am using React Native with Expo to be exact, I get the error when Expo is building the bundle.

jcchavezs:
Any chance any of you can come up with a fix for this?

santiagoalmeidabolannos:
I am not sure which of your modules are currently using Node standard library modules

jcchavezs:
Ping @jdb8

jdb8:
So if I understand correctly, this issue is happening because the code is being imported outside of a node environment: and my change introduced an import-time side effect which assumes the environment is node.

[...]

PR #524, opened 2021-01-27
Fix: Module not found: node-fetch

Raunow:
Since the fetch implementation is figured out at runtime as described [here](https://github.com/openzipkin/zipkin-js#typescript).
I'd like to add the `"browser"` field to the `package.json` too in order to fix bundlers like webpack.
You can see more about the [`"browser"` field here](https://github.com/defunctzombie/package-browser-field-spec).

In short, this is a better fix than modifying (in my case non-existent) `tsconfig.json`

image

### DIFF
--- a/packages/zipkin-transport-http/package.json
+++ b/packages/zipkin-transport-http/package.json
@@ -23,5 +23,8 @@
     "body-parser": "^1.15.2",
     "fetch-retry": "3.1.0",
     "noop-logger": "^0.1.1"
+  },
+  "browser": {
+    "node-fetch": false
   }
 }


jcchavezs:
Thanks @Raunow for the fix and the details. Any input on this @anuraaga?

jcchavezs:
Is this also related to https://github.com/openzipkin/zipkin-js/issues/465?

Raunow:
> Is this also related to #465?

As far as I can tell, it's not the same issue.
As long as you don't use the following functions:
```js
os.userInfo()
os.getPriority()
os.version()
os.setPriority(0)
```
No errors will be thrown by an up to date chromuim based browser.
And I didn't get any errors from webpack with `require('os')`, unlike `require('node-fetch')`.

For reference the Zipkin npm package only makes use of `os.networkInterfaces()`

anuraaga (approved):
Seems to make sense, thanks

jcchavezs:
Thanks @Raunow I will publish a release very soon but in case you want to try the package, you can use the canary release:

A new version of the package zipkin-transport-http (0.22.1-alpha.5) was published at 2021-01-28T14:14:12.467Z from
35.196.72.151. The shasum of this package was ec6aef17241f32dbf9f90a3b101f9db742496a0f.

PR #525, opened 2021-02-26
Fix: Return of createNoopTracer is void

Raunow:
The definition for `createNoobTracer()` in `index.d.ts` says that I should expect a void return.
But looking at the code that's obviously incorrect.

The hints that are created using index.d.ts
image

The source code for `createNoopTracer` that always return a Tracer
image

Currently a workaround for proper typings in TypeScript is:
```ts
return <unknown>createNoopTracer() as Tracer;
```

### DIFF
--- a/packages/zipkin/index.d.ts
+++ b/packages/zipkin/index.d.ts
@@ -86,7 +86,7 @@ declare namespace zipkin {
     toString(): string;
   }
 
-  const createNoopTracer: () => void;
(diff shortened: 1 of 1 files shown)


jcchavezs (approved):
Thank you!

jcchavezs:
I wonder @Raunow if there should be any static check to avoid such issues in the future?

PR #534, opened 2023-03-29
Fix bug where exceptions in other middleware are being swallowed

phelma:
In zipkin-instrumentation-koa, by calling `next().catch()` we catch any exceptions in the next middleware
if we are going to catch them we should throw again so that
they can be handled elsewhere.


codefromthecrypt:
is there a way we can update the tests so they fail without this fix?

